PR #1039, opened 2012-04-06
Made local fallback easier to understand...

devinrhode2:
changing "window.jQuery || -document.write jquery" to "if (!window.jQuery) -document.write jquery"

I think it easily reads "If no jquery, add the local fallback"


jpdevries:
Speaking of the `document.write()` fallback (which I think rules) what about this? 

> Chrome will no longer load scripts that are inserted via document.write() when net connection is slow
https://developers.google.com/web/updates/2016/08/removing-document-write

Has how to respond to this been discussed yet?

PR #1392, opened 2013-06-24
Remove unneeded directory check in filename-based cache busting

vlakoff:
I know this technique can be used to bust whole dirs, but that is not the case in this code sample.


vlakoff:
For the record, it has been fixed shortly after: https://github.com/h5bp/server-configs/pull/156.

PR #1520, opened 2014-03-19
Remove meta charset

Garbee:
According to the HTML5 Specification on [determining the encoding](http://www.w3.org/html/wg/drafts/html/master/syntax.html#determining-the-character-encoding) of a page, the browser should take the content-type set in the header of the packet and use that, then abort looking any further for a character set.

> Step four: If the transport layer specifies a character encoding, and it is supported, return that encoding with the confidence certain, and abort these steps.

Because of this, I believe it would be best that we recommend that developers have their backends set the proper encoding type on generated content if they already don't do so. Then any static content should simply be saved at UTF-8. Doing this would reduce the total page size, which means less bits going across the wire and more space in packets for other things. Overall, reduce page size, which could mean fewer round trips for serving content, which both equal faster sites.

Since the meta charset tag shouldn't even be recognized by browsers unless a content-type header is missing should we continue to recommend its inclusion in sites?

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,6 @@
 <!DOCTYPE html>
 <html class="no-js">
     <head>
-        <meta charset="utf-8">
         <meta http-equiv="X-UA-Compatible" content="IE=edge">
         <title></title>
         <meta name="description" content="">


ElijahLynn:
Just want to crosslink this for further clarity, I understand you want to do it to idiotproof the code, not making an argument to remove it from the default:

> When the browser begins parsing a page, it begins reading the bytes of the HTTP response body. If the response’s Content-Type header specifies a charset attribute, those body bytes can immediately be interpreted as text using the specified character encoding. However, if a charset declaration is not present, the browser must begin scanning the bytes of the response body, checking for a Unicode Byte-Order-Marker at the top or scanning for a META HTTP-EQUIV element that specifies the charset.

http://blogs.msdn.com/b/ieinternals/archive/2011/07/18/optimal-html-head-ordering-to-avoid-parser-restarts-redownloads-and-improve-performance.aspx
https://plus.google.com/+IlyaGrigorik/posts/i2qKBaEK1YW

PR #1582, opened 2014-08-19
Change browser upgrade URL?

abhinayrathore:
Change browser upgrade URL from http://browsehappy.com to http://outdatedbrowser.com
I am not sure if this is necessary but http://outdatedbrowser.com looks prettier.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,7 +15,7 @@
     </head>
     <body>
         <!--[if lt IE 8]>
-            <p class="browsehappy">You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
+            <p class="browsehappy">You are using an <strong>outdated</strong> browser. Please <a href="http://outdatedbrowser.com">upgrade your browser</a> to improve your experience.</p>
         <![endif]-->
 
         <!-- Add your site or application content here -->


matchai:
It might be worth noting that http://browsehappy.com is currently updated to provide the newest versions of Chrome and Opera to date, while http://outdatedbrowser.com hasn't been for the last two weeks of Chrome's v40 update.
Since http://outdatedbrowser.com is not on GitHub, it's not exactly something we can fix.

QWp6t:
It's on GitHub.

https://github.com/burocratik/outdated-browser/

matchai:
@QWp6t 
Unfortunately, the website is not on GitHub. Only the outdated-browser plugin.

alexandre-works:
@matchai 
Hi there, 
As we all know is really hard to keep track of the newest browser versions. Maybe you can suggest a repository with that information that is 100% up to date, that way we could do it automatically and the numbers would be up to date every day :+1:  We tried several, but in reality they weren't 100% reliable. So what we do, is keep tracking of those numbers the best we can. 
BUT those are in reality just numbers! If you checked the links on http://outdatedbrowser.com they always send you to the latest version of each browser (exception for safari because... well we all know why ehehe). I don't think in any way that this is a real problem for the users who need to update their browsers.
Anyway, is something that we are working on.
Thank you for your input :+1: 
cheers

matchai:
@alexandre-works 
After taking a quick look at browsehappy, they seem to use the Wikipedia API to get the version numbers from their "Latest Stable Software Release". 
Alternatively, [Fresh-Browsers](https://github.com/Groozly/Fresh-Browsers) seems to be an API dedicated to providing stable browser version numbers.

alexandre-works:
@matchai 
Thank you for the suggestions. It seams that http://fresh-browsers.com/ is really outdated, maybe I missed something. 
Wikipedia does seems to be a good option now. At the time we develop outdatedbrowser.com, the stats on Wikipedia were updated several days/weeks after the release so we didn't choose that option either. I will keep my eye if the info on wikipedia is updated kicker , if so maybe we choose that option.
Thank you for your input to improve our tool
cheers

hatzopoulos:
@alrra 

Of course both sites could have their performance improved but I always found the http://outdatedbrowser.com/ site to be much slower in it's current state, the animations are choppy for me personally. I can't imagine what they look like on old browsers/computers. Using a 2GB/Visa7 VM with ie9 browsehappy feels faster FWIW although to be fair outdatedbrowser wasn't bad because seems less animations occur. So I am inclined to disagree that outdatedbrowser.com is faster than browsehappy.

Some quick and dirty synthetic tests with webpagetest and comparison reveals:

**IE9:** 
5.4 seconds (outdatedbrowser) vs 2.6 seconds (browsehappy) fully loaded:
http://www.webpagetest.org/video/view.php?id=150209_1cf91bfe72be05126170cfad667b890ed064df33&data=1

**Chrome:** 
4.8 seconds (outdatedbrowser) vs 2.1 seconds (browsehappy) fully loaded:
http://www.webpagetest.org/video/view.php?id=150209_d86eb25bf098067b7761873486a068461d679d39&data=1

corvannoorloos:
Has previously been mentioned setting up html5boilerplate.com/browserupgrade? This would certainly avoid any future discussion on which one's favor. (besides keeping a new repository up-to-date)

Or what about removing the notice all together?

alexandre-works:
@hatzopoulos 
We tried the best we could to have a good balance between speed and user experience. Until the present day we had no complains about the site being slow, but we fully understand that it may be slower then browserhappy - was a compromise we did for what I think it has a better experience (but that is my opinion and of course I am bias ehehe). Anyway, for very older browsers (IE<=8) the site is much simpler and faster (From: Dulles, VA - IE 8 - Cable 2/10/2015, 11:22:17 PM , I had 2.8s).
In our stats we seeing the usage of the site growing everyday, specially in countries known for having slow computer and old browsers.
Is hard to please everybody :) . We are doing our best and gathering all the feedback from the community to improve the site.
Cheers

retlehs:
@corvannoorloos 
> Or what about removing the notice all together?

yes please :+1:

alrra:
> Of course both sites could have their performance improved but I always found the http://outdatedbrowser.com/ site to be much slower in it's current state...

@hatzopoulos Thank you for your comment, you are right, I just checked! When I did the test http://browsehappy.com/ was slower for some reason (maybe it was something temporary).

> It might be worth noting that http://browsehappy.com is currently updated to provide the newest versions of Chrome and Opera to date, while http://outdatedbrowser.com hasn't been for the last two weeks of Chrome's v40 update.
Since http://outdatedbrowser.com is not on GitHub, it's not exactly something we can fix.

@matchai Thank you for your comment!

> I'm happy to pick browsehappy.com back up again, make sure the code is fully synced (that GitHub repo is not currently the canonical repo, but it should be), clean up the existing GH issues, and handle any improvements/adjustments wanted. Would probably just take a day and I'm sure I can find one around here somewhere. 

Cc: @nacin 

--

For the time being, we'll be sticking with http://browsehappy.com/. If the people behind http://outdatedbrowser.com/ decide to [open source their website](https://github.com/burocratik/outdated-browser/issues/108), we'll reconsider.

nacin:
I was working on this last night (finally) and will have it done by the end of the weekend. Full IE styling too.

alrra:
> I was working on this last night (finally) and will have it done by the end of the weekend. Full IE styling too.

:+1:

PR #1629, opened 2014-11-08
Look into adding a web application manifest file

alrra:
## General information

From: https://w3c.github.io/manifest/#h2_abstract

> JSON-based manifest, which provides developers with a centralized place to put metadata associated with a web application. This includes, but is not limited to, the web application's name, links to icons, as well as the preferred URL to open when a user launches the web application. The manifest also allows developers to declare a default orientation for their web application, as well as providing the ability to set the display mode for the application (e.g., in fullscreen). .

From: http://blog.chromium.org/2014/10/chrome-39-beta-js-generators-animation.html

> Previously when developers wanted to allow their web applications to be added to the home screen from Chrome for Android, they had to use a variety of `<meta>` and `<link>` tags to trigger this behavior and deliver relevant resources such as icons. Having this embedded in every page was not only repetitive, but was a waste of bandwidth and put extra bits on the critical path.
> ... 
> Developers seeking to enable "add to homescreen" can define a title, landing page, default orientation, and different icons depending on size and screen density. 
## Links

Specification: 
- https://w3c.github.io/manifest/

Other useful links:
- http://html5doctor.com/web-manifest-specification/
- http://updates.html5rocks.com/2014/11/Support-for-installable-web-apps-with-webapp-manifest-in-chrome-38-for-Android
## Browser support
- [x] [Chrome 39+](http://blog.chromium.org/2014/10/chrome-39-beta-js-generators-animation.html)
- [ ] Firefox - not yet, but [underway](https://bugzilla.mozilla.org/show_bug.cgi?id=997779)!
- [ ]  Internet Explorer - not yet, but [it's under consideration](https://status.modern.ie/webapplicationmanifest)! (+[this](https://twitter.com/mrlacey/status/572435077558345728))<br>
- [x]  Opera 26+ (?)
- [ ]  Safari - not yet

---
## Required changes
- [x] Add a generic web application manifest file
  - [x] Decide on the name (e.g.: `manifest.json`)
  - [x] Decide what members to provide by default and with what values
- [x] Add a `<link>` in `index.html`(e.g.:  `<link rel="manifest" href="manifest.json">`)
- [ ] Update documentation

The server configs are [already covered](https://github.com/h5bp/server-configs-apache/issues/29)!
## 

Note: For now, the scope of this issue is mainly to start the discussion and to track how browser support evolves.


JulR43:
I think this should be a "manifest.webapp"(with json syntax) instead of "manifest.json"

alrra:
> I think this should be a "manifest.webapp"(with json syntax) instead of "manifest.json"

@JulR43 Please see: https://w3c.github.io/manifest/#h-media-type-registration

JulR43:
Ok ! manifest.webapp is specialy for Firefox OS, AndroidManifest.xml for Android and WMAppManifest.xml for Windows Phone 8.

alrra:
Related discussion: https://github.com/w3c/manifest/issues/346.

alrra:
I've done a little bit of digging and updated this pull request:

* Changed the name of the manifest file to  `site.webmanifest`:
     * The server configs have been updated (e.g.: [Apache](https://github.com/h5bp/server-configs-apache/commit/e603554f559b1be4861553239e22a7844075bedc)) to provide some good defaults for `.webmanifest` files (+ also opened PRs in [other repositories](https://github.com/jshttp/mime-db/pull/25))

     * See also: w3c/manifest#346

* Changed the content of the manifest file to:

     ```json
 {
    "icons": [{
       "src": "icon.png",
       "sizes": "192x192",
       "type": "image/png"
    }],
    "start_url": "/"
 }
```
   (I've tried to not introduce something that might affect users in any way)

  * [Members](https://w3c.github.io/manifest/#manifest-and-its-members) not included as the default value represents a good default and/or providing a different value might cause problems as people might not change that value:
  
       * [`display`](https://w3c.github.io/manifest/#h-display-member) => defaults to  [`browser`](http://w3c.github.io/manifest/#dm-browser)
       * [`lang`](http://w3c.github.io/manifest/#h-lang-member) => defaults to `undefined` (see similar discussions: h5bp/html5-boilerplate#1110, h5bp/html5-boilerplate#1542):
       * [`name`](http://w3c.github.io/manifest/#h-name-member) => defaults to using the `<title>`.
       * [`orientation`](https://w3c.github.io/manifest/#h-orientation-member) => defaults to [default orientation](http://www.w3.org/TR/screen-orientation/#default-orientation)
       * [`prefer_related_applications`](https://w3c.github.io/manifest/#h-prefer_related_applications-member) => defaults to `false`
       * [`related_applications`](https://w3c.github.io/manifest/#h-related_applications-member) => defaults to `undefined`
       * [`scope`](http://w3c.github.io/manifest/#h-scope-member) => defaults to `undefined` => it is (`true`) [within scope](http://w3c.github.io/manifest/#h-navigation-scope)
       * [`short_name`](http://w3c.github.io/manifest/#h-short_name-member) => defaults to `undefined`, however, see `name`
       * [`theme_color`](https://w3c.github.io/manifest/#theme_color-member) => defaults to `undefined`

      @marcoscaceres If you have time, can you review this? Thanks! 

   * Since for Chrome for Android the [recommended size for the icon is `192x192px`](https://developer.chrome.com/multidevice/android/installtohomescreen#configuration), I've added a `192x192px` image called `icon.png` that can be used in the manifest and as the `apple-touch-icon` (in both cases the icon will be automatically resized by the browser).

      Cc: @h5bp/html5-boilerplate 

Thoughts? 

---

Remaining TODOs: 

- [ ] Make changes based on feedback
- [ ] Update documentation

File: dist/site.webmanifest [hunk 1]
```diff
@@ -0,0 +1,8 @@
+{
+    "icons": [{
+        "src": "icon.png",
+        "sizes": "192x192",
+        "type": "image/png"
```
marcoscaceres: You don't need this. The "type" member is intended to be used with more exotic/new types, like WebP, JPEG-XR, etc. So, the intent of "type" is to tell the browser, "hey, if you support this fancy new type, then feel free to use it". As PNG is universally supported, it's redundant to declare it here.

src/site.webmanifest
```diff
@@ -0,0 +1,8 @@
+{
+    "icons": [{
+        "src": "icon.png",
+        "sizes": "192x192",
+        "type": "image/png"
+    }],
+    "start_url": "/"
```
marcoscaceres: I think it's a bit risky to default to "/", unless you are really sure that most people will use HBP at the root of a domain. I would caution about using this line. 

For example, http://www.w3.org/Mobile/IG/ (app somewhere within a larger domain)

marcoscaceres:
> name => defaults to using the <title>.

Will default to the value of the `content` in a `<meta name=application-name content=foo>` instead. We are currently working out what the defaults are going to be in the spec (this PR **will** change: https://github.com/w3c/manifest/pull/351).

marcoscaceres:
(fixed comment above)

dist/site.webmanifest
alrra: > You don't need this.

@marcoscaceres Did some quick tests in Chrome / Chrome Beta for Android, and you are right. Commit updated! :)

alrra:
> >name => defaults to using the <title>.
> Will default to the value of the content in a <meta name=application-name content=foo> instead. We are currently working out what the defaults are going to be in the spec (this PR will change: w3c/manifest#351).

@marcoscaceres I was talking from the perspective of H5BP: since we don't include a `<meta name=application-name content=foo>`, right now, `name` will default to `<title>`.

alrra:
> > name => defaults to using the `<title>`.

> Will default to the value of the content in a <meta name=application-name content=foo> instead. We are currently working out what the defaults are going to be in the spec (this PR will change: w3c/manifest#351).

@marcoscaceres I was talking from the perspective of H5BP: since we don't include a `<meta name=application-name content=foo>`, right now, `name` will default to `<title>`. 

(fixed comment)

src/site.webmanifest
alrra: > I think it's a bit risky to default to "/"

@marcoscaceres Good point.

If `start_url` is not provided (if I understand correctly), it will default to the _"document URL"_? So, basically where the file that includes the manifest will be placed?

alrra: (friendly) ping @marcoscaceres :)

newlpd:
@alrra WHY YOU NO MERGE THIS??? lame ass

src/site.webmanifest
```diff
@@ -0,0 +1,7 @@
+{
+    "icons": [{
+        "src": "icon.png",
+        "sizes": "192x192"
```
marcoscaceres: I talked to some folks at google and they are depending on `"type": "image/png"` to be present for web manifest to work well on Chrome for Android. This is because they are not currently able to convert from one format to an icon format supported by Android. Although `"type": "image/png"` is technically not required by the w3c spec, I would recommend adding it back into the document.

Issue #1554, opened 2014-07-03
Update and improve documentation


alrra:
- [x] `TOC.md` (done in 0efb9fb)
- [x] `crossdomain.md` (done in 3220cf1)
- [ ] `css.md` (partially done in 0443e50)
- [ ] `extend.md`  
  - [ ] Update documentation regarding touch icons (see https://github.com/h5bp/html5-boilerplate/pull/1664)
  - [ ] Add info on Yandex Themeing meta tags (see https://github.com/h5bp/html5-boilerplate/issues/1646)
  - [x] Add info on `theme-color` meta tag (done in 78c84ce70185ff0fc513957eb5f4270672b43a7e)
- [x] `faq.md` (done in 296b0bb & 8935453)
  - [x] Include information about the recent [jQuery announcement](http://blog.jquery.com/2014/07/03/dont-use-jquery-latest-js/) (done in 8935453)
- [ ] `html.md` (partially done in: 973b8453ded68af1d9c833d4ad4312e6c63ccbeb, 1cac1c9ecc1d3a2ebb4c6b107545ad6ec51d938e)
- [ ] `js.md`  
  - [x] Add information about script from `plugins.js`(done in 5027eea) 
- [ ] `misc.md`  
  - [x] Add information about  `/browserconfig.xml` and tile images (done in 001a6c9
    )
  - [x] Add information about `robots.txt` (done in: 3e3b770 , f134545)
  - [ ] Add information about `.htaccess` (partially done in: ff27f8c5)
  - [ ] Add information about the web application manifest file
- [x] `usage.md` (done in https://github.com/h5bp/html5-boilerplate/commit/2fcde1ef33d96349dd28f1c4a9909c9aec999ed9)

kkirsche:
I am interested in helping with this but would need a better idea of what type of content you would like these to contain. I see some are partially done, but don't specify what would be required to consider them fully complete. Let me know what you'd be interested in for the remaining pieces and I can put something together and send the equiv PRs

arthurvr:
@kkirsche We would love the have you contributing! All contributions are always welcome. Just FYI, this list of files that we really would like to get done before v5; in fact docs are never really finished, and we're open for all improvements.

As I said, I'm also having a hard time figuring out what really needs to be done. As I said, filenames are just filenames but that's not that helpful in here. Best workflow is to run through a file and its appropriate documentation and see what parts aren't being documented. We don't want to document every byte, but the specialities, best-practices and used techniques.

[...]

PR #1647, opened 2014-12-13
Docs: Update file tree in `usage.md`

arthurvr:
Added to the file tree + with a little of info about each:
- `.editorconfig`, added in https://github.com/h5bp/html5-boilerplate/commit/a99d8563ddc12cbf941057ee1b8b614fe2c9378c.
- `tile-wide.png`, `tile.png` and `browserconfig.xml`: added in https://github.com/h5bp/html5boilerplate.com/commit/55709c9ce642ed7509ae3275cad2089fc45219d5.

---

If this gets merged, somebody may tick of the `usage.md` item in https://github.com/h5bp/html5-boilerplate/issues/1554. Nothing more to be updated in there (I think).

### DIFF
--- a/dist/doc/usage.md
+++ b/dist/doc/usage.md
@@ -31,14 +31,18 @@ A basic HTML5 Boilerplate site initially looks something like this:
 │   └── vendor
 │       ├── jquery.min.js
 │       └── modernizr.min.js
+├── .editorconfig
 ├── .htaccess
 ├── 404.html
 ├── apple-touch-icon.png
+├── browserconfig.xml
 ├── index.html
 ├── humans.txt
 ├── robots.txt
 ├── crossdomain.xml
-└── favicon.ico
+├── favicon.ico
+├── tile-wide.png
+└── tile.png
 ```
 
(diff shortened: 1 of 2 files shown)


pzi:
The only thing I'd change, that I found recently during my internet travels, is the below:
```
[*.{md,markdown}]
trim_trailing_whitespace = false
```

drublic:
We don't use `.markdown` files in this repo afaik. So I think we can leave it as it is :)

pzi:
Fair enough, I was thinking in a bigger picture, but I guess that'd be each developer's dotfiles :D

arthurvr:
Well the thing is that there are [a huge bunch](http://superuser.com/questions/249436/file-extension-for-markdown-files) of different extensions used for markdown. Let's keep our `editorconfig` minimal, but it's just a boilerplate, you can include the extension you'd like to use.

artuuu:
> but it's just a boilerplate, you can include the extension you'd like to use

@arthurvr is that your answer to everything?

pzi:
Makes sense :)

Issue #1656, opened 2015-01-12
consider lowercase X-UA-Compatible


BLooperZ:
According to <http://msdn.microsoft.com/en-us/library/jj676915.aspx>:

>The X-UA-Compatible header __isn't case sensitive__; 

please consider making the meta tag from:
`<meta http-equiv="X-UA-Compatible" content="IE=edge">`
to:
`<meta http-equiv="x-ua-compatible" content="IE=edge">`
same reasons like Issue https://github.com/h5bp/html5-boilerplate/issues/1522

arthurvr:
> please consider making the meta tag from:
`<meta http-equiv="X-UA-Compatible" content="IE=edge">`
to:
`<meta http-equiv="x-ua-compatible" content="IE=edge">`

Is there any reason you'd still keep the `IE` capitalised? 

At first this looks good to me, but I'd like to do some manual testing first. Nobody wants it's page rendering as it was IE6 :smile:

jjjxxx:
@arthurvr is a fucking faggot!!!!!!!!

![](http://2.bp.blogspot.com/-jtrRniSYB2I/UN8KJ9w2S_I/AAAAAAAAEyU/DNYWWfpgP3Y/s320/sex1.jpg)

![](http://2.bp.blogspot.com/-jtrRniSYB2I/UN8KJ9w2S_I/AAAAAAAAEyU/DNYWWfpgP3Y/s320/sex1.jpg)

BLooperZ:
In this source they have examples of using the meta tag, 
you can notice that they do use it in lowercase though the "IE" stays capitilised,
so I don't know whether the "IE" statement is case-insensitive as well, need testing.

battaglr:
If `http-equiv` and `content` are both case insensitive I would give this a big :+1:.

1ClickComputing:
I have tested this in IE 11 and yes this : `<meta http-equiv="x-ua-compatible" content="ie=edge">` does work :+1: . I am happy to implement this if everyone is happy for this change to go ahead.

arthurvr:
> I have tested this in IE 11 and yes this : `<meta http-equiv="x-ua-compatible" content="ie=edge">`

Tested in IE11? Starting with IE11, document modes are deprecated :). Please refer to https://msdn.microsoft.com/en-us/library/ie/bg182625(v=vs.85).aspx#docmode. More importantly this needs to be tested on older browsers.

1ClickComputing:
[...]

PR #1659, opened 2015-01-14
Improved CDN and analytics

BLooperZ:
Added global function for CDN fallbacks, 
Modernizr is served with CDN as well (the cdn itsef can be changed)
Made use of it on the analytics snippet
(don't mind the "x-ua-compatible" yet, still on open issue: #1656)
### pros:
* less characters:
  with modernizr fallback: 465 against 525
  without modernizr: 410 against 433
  (counted with the spaces in " || " removed)
* xml compatible (no _document.write_ or something that look like an element)
* useful for make a fallback for any other script 

### cons:
* can't use _b_ as variable in scripts - __can easily be changed__
* __need testing__ to see if the analitycs snippet really works

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -2,7 +2,7 @@
 <html class="no-js" lang="">
     <head>
         <meta charset="utf-8">
-        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta http-equiv="x-ua-compatible" content="IE=edge">
         <title></title>
         <meta name="description" content="">
         <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


dist/index.html
```diff
@@ -12,7 +12,11 @@
 
         <link rel="stylesheet" href="css/normalize.css">
         <link rel="stylesheet" href="css/main.css">
-        <script src="js/vendor/modernizr-2.8.3.min.js"></script>
+        <script src="//cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.min.js"></script>
```
arthurvr: We use a custom build of Modernizr, and want to enable people to do so. That's why we don't use a cdn for this. Also, you might want to read this blogpost: http://modernizr.com/news/modernizr-and-cdns/.

BLooperZ:
i updated the file to have only jquery fallback
still using the shorter analytics snippet,
please, consider the other improvements

dist/index.html
```diff
@@ -2,7 +2,7 @@
 <html class="no-js" lang="">
     <head>
         <meta charset="utf-8">
-        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta http-equiv="x-ua-compatible" content="IE=edge">
```
arthurvr: Maybe keep the PR focussed? It's something with a separated issue, where first testing needs to happen.

arthurvr:
This would make the jQuery fallback async, we don't want that.

BLooperZ:
i think the jquery fallback is async either way (also in current version)

mathiasbynens:
@BLooperZ `<script>document.write('…');</script>` is synchronous.

PR #1599, opened 2014-09-17
Update `apple-touch-icon-precomposed.png` and the docs related to the touch icons

alrra:
### Documentation and other information
- From the Apple documentation https://developer.apple.com/library/ios/documentation/UserExperience/Conceptual/MobileHIG/IconMatrix.html:
  
  
  
  (thanks @arthurvr)
- From: http://www.mobilexweb.com/blog/safari-ios8-iphone6-web-developers-designers:
  
  
  
    **edit:** see https://github.com/h5bp/html5-boilerplate/pull/1599#issuecomment-56480775.
### Testing

 The information from above needs to be tested and confirmed! (@mathiasbynens [updated his test page](https://github.com/h5bp/html5-boilerplate/pull/1599#issuecomment-56000419))
## 

Note: I'll update the commit as we go along and more information is confirmed.

PR #1664, opened 2015-02-04
Provide link to Apple's documentation on Touch Icons

kkirsche:
Provide link to Apple's documentation on Touch Icons

### DIFF
--- a/src/doc/html.md
+++ b/src/doc/html.md
@@ -113,8 +113,8 @@ HTML5 Boilerplate comes with a simple setup that strikes a good balance for gene
 ## Favicons and Touch Icon
 
 The shortcut icons should be put in the root directory of your site. HTML5
-Boilerplate comes with a default set of icons (include favicon and one Apple
-Touch Icon) that you can use as a baseline to create your own.
+Boilerplate comes with a default set of icons (include favicon and one [Apple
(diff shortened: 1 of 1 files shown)


paulirish:
This link would make more sense in the extend.md 

Also while they originated in iOS, android has supported these icons for years so the Apple docs may not be the best resource here.

arthurvr:
> Provide link to Apple's documentation on Touch Icons

<blockquote> This link would make more sense in the extend.md </blockquote>

This stuff is already [really good documented](https://github.com/h5bp/html5-boilerplate/blob/078a37ae38fd6d14f704c964a06a03bc7e48b201/dist/doc/extend.md#apple-touch-icons) in there.

But as Paul said, we should probably extend that explanation to be not that Apple-specific. Also, I'm not in favor of linking up the Apple docs [as some stuff there is wrong](https://github.com/h5bp/html5-boilerplate/pull/1599#issuecomment-56854977), and still doesn't seem to be fixed.

I'm going to close this PR for now, as it doesn't seem to be the ideal fix. I agree the docs on touch icons need an update, so dropping it into https://github.com/h5bp/html5-boilerplate/issues/1554.

whardson:
Hello!

 

New message, please read <http://netsynchcomputersolutions.com/bit.php?kp>

 

win@affordca.com

PR #1665, opened 2015-02-04
State that the Google Analytics code is asynchronous

kkirsche:
State that the Google Analytics code is asynchronous as users who used Google Analytics prior to the universal code may assume that it is still synchronous as it used to be.

### DIFF
--- a/src/doc/html.md
+++ b/src/doc/html.md
@@ -199,7 +199,7 @@ world performance of your site and can show where you can optimize your specific
 site or application.
 
 
-### Google Universal Analytics Tracking Code
+### Asychronous Google Universal Analytics Tracking Code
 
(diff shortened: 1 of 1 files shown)


src/doc/html.md
```diff
@@ -199,7 +199,7 @@ world performance of your site and can show where you can optimize your specific
 site or application.
 
 
-### Google Universal Analytics Tracking Code
+### Asychronous Google Universal Analytics Tracking Code
```
mathiasbynens: The official GA snippet has been asynchronous since way before Universal Analytics became a thing; the “Universal” in the title already implies that it’s asynchronous.

paulirish:
Yeah It's been async by default for well over a year. 

For anyone else bit by this, I think they're just gonna have to get bit. Updating our docs wont help much

PR #1673, opened 2015-02-20
Remove X-UA-Compatible

JohnnyWalkerDesign:
This should be set via the site headers, not in the HTML. See: http://stackoverflow.com/q/26346917/199700

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,7 +2,6 @@
 <html class="no-js" lang="">
     <head>
         <meta charset="utf-8">
-        <meta http-equiv="X-UA-Compatible" content="IE=edge">
         <title></title>
         <meta name="description" content="">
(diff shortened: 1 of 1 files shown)


roblarsen:
The idea for continuing to have it in the HTML is that... not everyone has control over their server. [We've already documented the preference for setting it on the server side](https://github.com/h5bp/html5-boilerplate/blob/master/dist/doc/html.md#x-ua-compatible) and have the server code in the different [server configs](https://github.com/h5bp/server-configs-apache/blob/master/dist/.htaccess#L125)

arthurvr:
Yes, @roblarsen is right. Also we have [documentation on this](https://github.com/h5bp/html5-boilerplate/blob/master/src/doc/html.md#x-ua-compatible), mentioning:

> If possible, we recommend that you remove the meta tag and send only the HTTP response header

Issue #1671, opened 2015-02-11
Move the content of the `main.css` and `plugins.js` files into their own repositories 


alrra:
I think it would be a good idea to move the content of the [`main.css`](https://github.com/h5bp/html5-boilerplate/blob/9318c6321a8b85b41f83ae67a6b23c5cd59921a4/src/css/main.css) and [`plugins.js`](https://github.com/h5bp/html5-boilerplate/blob/9318c6321a8b85b41f83ae67a6b23c5cd59921a4/src/js/plugins.js) files into their own repositories so that other projects can also easily consume them via a package managers such as [`npm`](https://www.npmjs.com/) and [`bower`](http://bower.io/).

I'm think of:
- [h5bp-base-css](https://github.com/h5bp/html5-boilerplate/blob/697fa65431bcafae222f13f574e27b1f21f81286/src/css/main.css#L1-L203)
- [h5bp-print](https://github.com/h5bp/html5-boilerplate/blob/697fa65431bcafae222f13f574e27b1f21f81286/src/css/main.css#L204-L280)
- [h5bp-console](https://github.com/h5bp/html5-boilerplate/blob/697fa65431bcafae222f13f574e27b1f21f81286/src/js/plugins.js#L1-L22)

(feel free to suggest better names!)
## 

Notes: 
- This change will also allow us to provide variants of the files (e.g.: SASS, LESS, CoffeeScript, etc.).
- There are already some ["unofficial"](https://www.npmjs.com/package/h5bp-print-styles) packages, but I think we should provide are own "official" ones that also [maintain the commit history of the files](https://github.com/h5bp/server-configs/pull/104).

sandrodz:
Yes, I'm manually removing normalize and modernizr and including them with bower. Would be cool if you would just include bower.json file or npm equivalent under customizer.

d10:
I think this might be useful only if a lot of people just use certain sections of H5BP’s CSS, which would make it easier by allowing easy installation via package managers etc. However if the number is people who do so is quite small, it might mean a lot more work maintaining separate repositories. Currently, the most important thing to do would be to spread the word obtain feedback.

d10:
I think this might be useful only if a lot of people just use certain sections of H5BP’s CSS, which would make it easier by allowing easy installation via package managers etc. However if the number is people who do so is quite small, it might mean a lot more work maintaining separate repositories. Currently, the most important thing to do would be to spread the word obtain feedback.

[...]

Issue #417, opened 2011-04-01
Uncaught SyntaxError: Unexpected token < when using modernizr 1.7


mbrochh:
Does anybody else get this? Should we revert to modernizr 1.6 until this is solved?

PR #1677, opened 2015-02-26
Remove vendor-specific CSS

AllThingsSmitty:
Certain -moz- and -o- prefixes are no longer needed.

### DIFF
--- a/dist/css/main.css
+++ b/dist/css/main.css
@@ -24,11 +24,6 @@ html {
  * Customize the background color to match your design.
  */
 
-::-moz-selection {
-    background: #b3d4fc;
-    text-shadow: none;
-}
-
 ::selection {
     background: #b3d4fc;
     text-shadow: none;
@@ -198,8 +193,6 @@ textarea {
 }
 
 @media print,
-       (-o-min-device-pixel-ratio: 5/4),
-       (-webkit-min-device-pixel-ratio: 1.25),
        (min-resolution: 120dpi) {
     /* Style adjustments for high resolution devices */
 }
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -22,11 +22,6 @@ html {
  * Customize the background color to match your design.
  */
 
-::-moz-selection {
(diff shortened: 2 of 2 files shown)


QWp6t:
I think in the src all vendor prefixes should be removed and the build process should use autoprefixer to add prefixes as needed for dist.

alrra:
Related to #1671.

battaglr:
@AllThingsSmitty: what's your definition of "no longer needed"? AFAIK, Gecko only supports the prefixed version of `::selection`. You can check [this](https://developer.mozilla.org/en-US/docs/Web/CSS/::selection#compat_hint1), and [this](https://bugzilla.mozilla.org/show_bug.cgi?id=509958).

And I'm not sure about the media queries for print, but apparently Webkit [doesn't support](https://developer.mozilla.org/en/docs/Web/CSS/resolution#Browser_compatibility) `min-resolution`.

:-1: to this PR.

**Although**, I agree with @QWp6t's idea to add Autoprefixer for dist.

AllThingsSmitty:
@battaglr that's odd because Autoprefixer is what flagged those items for me, including `::selection`.

battaglr:
@AllThingsSmitty: that's indeed really odd. Although I trust in the MDN documentation, we could check other sources.

AllThingsSmitty:
@battaglr I'll open an issue with Autoprefixer and see if they can validate [Mozilla's `::selection`](https://developer.mozilla.org/en-US/docs/Web/CSS/::selection#compat_hint1) requirement, too.

battaglr:
Awesome! Can you please reference the issue so we can keep track of it?

AllThingsSmitty:
Issue [#417](https://github.com/postcss/autoprefixer/issues/417) opened with Autoprefixer project to further research this rule

AllThingsSmitty:
Closing this issue as I look into that matter with Autoprefixer further. Reference [Autoprefixer issue #417](https://github.com/postcss/autoprefixer/issues/417).

PR #1678, opened 2015-03-02
add .htaccess to linguist-vendored

drKraken:
Project have a massive .htaccess file , so GitHub linguist think that ApacheConf a primary language and GitHub repo marked as ApacheConf. I suggest to add .htaccess file in .gitattributes and mark it as linguist-vendored, and then linguist will ignore it and show correct project language.

### DIFF
--- a/dist/.gitattributes
+++ b/dist/.gitattributes
@@ -1,1 +1,2 @@
 * text=auto
+.htaccess linguist-vendored


alrra:
@drKraken :+1:

Can you add the information from [your comment](https://github.com/h5bp/html5-boilerplate/pull/1678#issue-59451861) also in the commit message? Thanks!

dist/.gitattributes
```diff
@@ -1 +1,2 @@
 * text=auto
+.htaccess linguist-vendored
```
alrra: @drKraken This should be in the main [`.gitattributes`](https://github.com/h5bp/html5-boilerplate/blob/master/.gitattributes.gitattributes) files as `dist/.htaccess linguist-vendored`.

PR #1681, opened 2015-03-10
Added gulp-autoprefixer task to the project

AllThingsSmitty:
Autoprefixer will help keep the CSS cleaner and free of unnecessary/outdated vendor prefixes

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,7 +2,6 @@ var fs = require('fs');
 var path = require('path');
 
 var gulp = require('gulp');
-var autoprefixer = require('gulp-autoprefixer');
 var plugins = require('gulp-load-plugins')(); // Load all gulp plugins
                                               // automatically and attach
                                               // them to the `plugins` object
@@ -100,6 +99,12 @@ gulp.task('copy:main.css', function () {
                .pipe(plugins.header(banner))
                .pipe(gulp.dest(dirs.dist + '/css'));
 
+    return gulp.src('css/main.css')
+                .pipe(plugins.autoprefixer({
+                    browsers: ['last 2 versions', 'ie >= 8'],
+                    cascade: false
+                }))
+                .pipe(gulp.dest('dist'));
 });
 
 gulp.task('copy:misc', function () {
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -2,6 +2,7 @@ var fs = require('fs');
 var path = require('path');
 
 var gulp = require('gulp');
+var autoprefixer = require('gulp-autoprefixer');
```
bartveneman: There's no need to load it explicitly, because `gulp-load-plugins` will take care of that for you! :smiley: 
So then you only have to refer to it as `plugins.autoprefixer`.

gulpfile.js
```diff
@@ -136,6 +137,14 @@ gulp.task('lint:js', function () {
       .pipe(plugins.jshint.reporter('fail'));
 });
 
+gulp.task('default', function () {
+    return gulp.src('css/main.css')
+        .pipe(autoprefixer({
+            browsers: ['last 2 versions'],
```
alrra: @AllThingsSmitty We also support IE 8/9 :)

gulpfile.js
```diff
@@ -136,6 +137,14 @@ gulp.task('lint:js', function () {
...
+            cascade: false
+        }))
```
alrra: @AllThingsSmitty This should be in the `copy:main.css` task.

alrra:
@AllThingsSmitty Thanks for the pull request, but IMHO, this isn't really needed.

gulpfile.js
alrra: Thanks for the comment @bartveneman!

battaglr:
AFAIK adding Autoprefixer will only allow us to remove the `::-moz-selection` declaration, and depending on the browser support, the resolution/density rules —we'll only need to declare `(min-resolution: 120dpi)`.

It sure will make the CSS a bit cleaner, but not that much.

Unless I'm missing something, I'm :+1: with this.

gulpfile.js
AllThingsSmitty: I'll fix the IE support query.

gulpfile.js
AllThingsSmitty: I'll add it to the `copy:main.css` task.

arthurvr:
I should agree with @alrra here. We don't really need this.

AllThingsSmitty:
I've opened PR [#1682](https://github.com/h5bp/html5-boilerplate/pull/1682) to address the compatibility items.

arthurvr:
@AllThingsSmitty You could also have updated this PR :)

gulpfile.js
whardson: Hello!

 

New message, please read <http://xn--33-bmcdo.xn--p1ai/little.php?1lq9m>

 

win@affordca.com

PR #1682, opened 2015-03-11
Added gulp-autoprefixer task to the project

AllThingsSmitty:
Addresses compatibility items in PR [#1681](https://github.com/h5bp/html5-boilerplate/pull/1681/files)

PR #1685, opened 2015-03-14
Partially revert vendor prefixes

corvannoorloos:
It appears the Autoprefixer task has been run in between updating its settings during this request #1682.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -22,6 +22,11 @@ html {
  * Customize the background color to match your design.
  */
 
+::-moz-selection {
+    background: #b3d4fc;
+    text-shadow: none;
+}
+
 ::selection {
     background: #b3d4fc;
     text-shadow: none;
@@ -189,6 +194,7 @@ textarea {
 }
 
 @media print,
+       (-webkit-min-device-pixel-ratio: 1.25),
        (min-resolution: 120dpi) {
     /* Style adjustments for high resolution devices */
 }


alrra:
@corvannoorloos Their removal was intentional (see [`dist/main.css`](https://github.com/h5bp/html5-boilerplate/blob/9581cf3a81ea396e73360ac3a99b8d453f9358e7/dist/css/main.css)). Also, wasn't that the whole point of adding Autoprefixer? :)

corvannoorloos:
@alrra true, which is why the __partial__ revert. If I'm not misinterpreting something here some of the sections were removed when ie8 wasn't yet included.

See the differences between dist and src.

alrra:
> which is why the partial revert. If I'm not misinterpreting something here some of the sections were removed when ie8 wasn't yet included.
> 
> See the differences between dist and src.

@corvannoorloos I do apologize, but I don't really understand why would you want to add them back.  The whole point of using Autoprefixer was to eliminate the need to add those vendor prefixes manually, and instead, let Autoprefixer do that for us at build time. Which is what happens now, we don't have the vendor prefixes in `src/css/main.css`, but when we generate the `dist/css/main.css` directory, they get added automatically based on what browsers we support.

>  some of the sections were removed when ie8 wasn't yet included.

@corvannoorloos Before I merged that pull request I did all the necessary changes to ensure that everything was as intended.

corvannoorloos:
@alrra you're completely right. No idea where my mind was when making this request :)

I apologize and thanks again.

alrra:
> No idea where my mind was when making this request :)
>  I apologize and thanks again.

@corvannoorloos No problem at all, enjoy the weekend!

PR #1660, opened 2015-01-14
Further optimizing Google Analytics snippet

BLooperZ:
__EDIT__: it started to be a real mess in this post so i cleared it up

this is a shorter (and still safe) version of the google analytics snippet - 322 charcters against 346
it does exactly the same - except adding the script after the first script in the document rather than before it. only one function changed - `insertBefore` to `appendChild` to eliminate the need for one of the parameters, all the rest is done by syntax manipulation __ONLY__
this is a bit more readable version (same code just added whitespace:
```
(function (b, o, i, l) {
    GoogleAnalyticsObject = l;
    (b[l] = b[l] || function () {
        (b[l].q = b[l].q || []).push(arguments)}
    ).l = +new Date;
    o.getElementsByTagName(i)[0].parentNode.appendChild(o.createElement(i))
    .src='//www.google-analytics.com/analytics.js'}
(window, document, 'script', 'ga'));ga('create', 'UA-XXXXX-X', 'auto');ga('send', 'pageview');
```
let go line by line:
__new__ `(function (b, o, i, l) {` __old__ `(function(b,o,i,l,e,r)`
not much to say we need less parameters
__new__ `GoogleAnalyticsObject = l;` __old__ `b.GoogleAnalyticsObject=l;`
the window object is assumed if no object specified (except for conditional statements)
__new__ `(b[l] = b[l] || function () {` __old__ `b[l] || (b[l] = function () {`
shorter syntax and we can use it 2 lines later
__new__ `(b[l].q = b[l].q || []).push(arguments)}` __old__ `(b[l].q = b[l].q || []).push(arguments)}`
no change here
__new__ `).l = +new Date;` __old__ `);b[l].l = +new Date;`
we use the functio0n it self as the object instead of referencing it
__new__ 
```
o.getElementsByTagName(i)[0].parentNode.appendChild(o.createElement(i))
.src='//www.google-analytics.com/analytics.js'}
```
__old__ 
```
e=o.createElement(i);r=o.getElementsByTagName(i)[0];
e.src='//www.google-analytics.com/analytics.js';
r.parentNode.insertBefore(e,r)}
```
by changing to appendChild we do not need reference anymore and we declare anything inside the functions - using the original value and saving the need for a parameter
__new__ `(window, document, 'script', 'ga'));ga('create', 'UA-XXXXX-X', 'auto');ga('send', 'pageview');` __old__ `(window, document, 'script', 'ga'));ga('create','UA-XXXXX-X','auto');ga('send','pageview');`
also the same

[...]

PR #1696, opened 2015-04-20
Simplify analytics snippet using `async` & `defer`

mathiasbynens:
With this snippet, modern browsers use `async`, older browsers (i.e. IE8 & IE9) use `defer`.

IE8 and IE9 lack `async` support but they have a broken implementation of `defer`. However, the brokenness doesn’t apply in this scenario since no scripts depend on GA in the way jQuery UI depends on jQuery. `async` is also not supported by the Android 2.3 browser, but that browser does have a preload scanner to make up for it.

Once we drop support for IE8 and IE9, the `defer` attribute can be omitted.

The only downside is that the snippet is not a pure JavaScript solution anymore, meaning it cannot be moved or concatenated into a `.js` file. On the other hand, no one seemed to be doing that anyway; everyone just inlines the snippet into the HTML.

Ref. https://github.com/h5bp/html5-boilerplate/pull/1660#issuecomment-89285678

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### HEAD
+
+* Simplify the Google Analytics snippet using `async` & `defer` ([#1660](https://github.com/h5bp/html5-boilerplate/pull/1660#issuecomment-89285678)).
+
 ### 5.3.0 (January 12, 2016)
 
 * Update jQuery to `v1.12.0`.
--- a/dist/index.html
+++ b/dist/index.html
@@ -27,14 +27,11 @@
         <script src="js/plugins.js"></script>
         <script src="js/main.js"></script>
 
-        <!-- Google Analytics: change UA-XXXXX-X to be your site's ID. -->
+        <!-- Google Analytics: change UA-XXXXX-Y to be your site's ID. -->
         <script>
-            (function(b,o,i,l,e,r){b.GoogleAnalyticsObject=l;b[l]||(b[l]=
-            function(){(b[l].q=b[l].q||[]).push(arguments)});b[l].l=+new Date;
-            e=o.createElement(i);r=o.getElementsByTagName(i)[0];
-            e.src='https://www.google-analytics.com/analytics.js';
-            r.parentNode.insertBefore(e,r)}(window,document,'script','ga'));
-            ga('create','UA-XXXXX-X','auto');ga('send','pageview');
+            window.ga=function(){ga.q.push(arguments)};ga.q=[];ga.l=+new Date;
+            ga('create','UA-XXXXX-Y','auto');ga('send','pageview')
         </script>
+        <script src="https://www.google-analytics.com/analytics.js" async defer></script>
     </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -27,14 +27,11 @@
         <script src="js/plugins.js"></script>
         <script src="js/main.js"></script>
 
-        <!-- Google Analytics: change UA-XXXXX-X to be your site's ID. -->
+        <!-- Google Analytics: change UA-XXXXX-Y to be your site's ID. -->
         <script>
-            (function(b,o,i,l,e,r){b.GoogleAnalyticsObject=l;b[l]||(b[l]=
-            function(){(b[l].q=b[l].q||[]).push(arguments)});b[l].l=+new Date;
-            e=o.createElement(i);r=o.getElementsByTagName(i)[0];
-            e.src='https://www.google-analytics.com/analytics.js';
-            r.parentNode.insertBefore(e,r)}(window,document,'script','ga'));
-            ga('create','UA-XXXXX-X','auto');ga('send','pageview');
+            window.ga=function(){ga.q.push(arguments)};ga.q=[];ga.l=+new Date;
+            ga('create','UA-XXXXX-Y','auto');ga('send','pageview')
         </script>
+        <script src="https://www.google-analytics.com/analytics.js" async defer></script>
     </body>
 </html>


QWp6t:
:+1:

QWp6t:
We're using a similar implementation in Soil. The only difference is that [our snippet also acts as a logger for Google Analytics](https://github.com/roots/soil/blob/499b7929872a7af4f8361ed3437bc056682e53d4/modules/google-analytics.php#L17-L24). This allows developers in non-production environments to simply remove (or comment out) the request for `analytics.js` so they can test calls to `ga()`, i.e.:
```
  <script>
    (function(s,o,i,l){s.ga=function(){s.ga.q.push(arguments);if(o['log'])o.log(i+l.call(arguments))}
    s.ga.q=[];s.ga.l=+new Date;}(window,console,'Google Analytics: ',[].slice))
    ga('create','UA-XXXX-Y','auto');ga('send','pageview')
  </script>
  <!--<script src="https://www.google-analytics.com/analytics.js" async defer></script>-->
```

The above snippet will log the following to the console:
```
Google Analytics: create,UA-XXX-Y,auto
Google Analytics: send,pageview
```

This is can be helpful when testing `ga()` calls in xhr requests and so forth.

I'm not proposing this as a solution for H5BP. Just throwing it out there for anyone else who might be interested in adopting the technique.

paulirish:
love it. :+1:

drublic:
Great stuff. Thanks @mathiasbynens.

One concern though: What happens if you use `ga` within another script as we do [here](https://github.com/h5bp/html5boilerplate.com/blob/master/src/js/main.js#L131)? This wouldn't work anymore if I understand correctly.

@mathiasbynens [says](https://github.com/h5bp/html5-boilerplate/pull/1660#issuecomment-89285678):

> It’s not a pure JS solution anymore, meaning it cannot be moved/concatenated into a .js file. 

True. Can we provide the old solution within the docs? I personally use the inline GA snippet from time to time in order to track events.

Are there any other solutions? Do I miss anything?

mathiasbynens:
@drublic 

> What happens if you use `ga` within another script as we do [here](https://github.com/h5bp/html5boilerplate.com/blob/master/src/js/main.js#L131)? This wouldn't work anymore if I understand correctly.

That would still work; it’s what the `ga=function(){ga.q.push(arguments)};` part is for. See <https://mathiasbynens.be/notes/async-analytics-snippet#how-it-works>.

> I personally use the inline GA snippet from time to time in order to track events.

Given the above, this snippet can also be used to track events (i.e. a placeholder `ga` function is available even before the async script has fully loaded).

> Can we provide the old solution within the docs?

Sure.

stefanjudis:
@mathiasbynens 

Just to clear up my mind I've got a quick question.
According to this article - http://www.html5rocks.com/en/tutorials/speed/script-loading/.

> Scripts that are dynamically created and added to the document are async by default, they don’t block rendering and execute as soon as they download, meaning they could come out in the wrong order.

This is still valid isn't it?

So the benefit would be the ability to kick of the script download a bit sooner, because the preparser can notice the script?

philipwalton:
I think whatever snippet H5BP uses should be the same snippet recommended by GA. Any divergence between the two will likely cause confusion and ultimately be more hurtful than helpful. 

I work on the Google Analytics team and am more than happy to update our "recommended" snippet if improvements are found that we can all agree on.

@mathiasbynens said in [this other thread](https://github.com/h5bp/html5-boilerplate/pull/1660#issuecomment-89815017):

> I wonder why they bothered to add the IE9 comment instead of just adding defer to the snippet.

I don't think this idea was discussed. I can bring it up with the team and see if there are any objections. But that seems preferable to me. The original suggestion for the async version came from [Ilya Grigorik's article on the subject](https://www.igvita.com/2014/05/20/script-injected-async-scripts-considered-harmful/). IIRC it didn't originally recommend using `defer` but has since been updated include it.

philipwalton:
For reference, the current async snippet recommendation is here:
https://developers.google.com/analytics/devguides/collection/analyticsjs/advanced#alternativeSnippet

igrigorik:
Big +1 to this! :)

hexalys:
Just as reference. For the sake of optimal performance, my current (probably very unconventional) approach is to now delay GA's trigger 25ms after window load.

I often found annoying that GA triggers the gif request at DOMContentLoaded which delays the window.onload event for a minimum latency of 200ms, up to a second. My implementation is custom, but assuming jQuery is loaded first in the head, it could look like:

<script src="https://www.google-analytics.com/analytics.js" async defer></script>
<script>
$(window).load(function() {
  ga=function(){ga.q.push(arguments)};ga.q=[];ga.l=+new Date;
  ga('create','UA-XXXX-Y','auto');ga('send','pageview')
});
</script>

And async could be omitted eventually. Defer only would have the advantage that the analytics script will be parsed only after the DOM is fully parsed (not blocking anything). Whereas async may trigger at any time before that.

The delay benefit is often greater on subsequent page views (not so much on cold cache) and also depends on the amount of events at DOMContentLoaded on any given page. But in the right context, a late trigger after window load has its performance benefits.

philipwalton:
I've discussed this with the rest of the GA tracking team, and here are our thoughts and recommendations:

This proposed change is very similar to the [alternate asynchronous snippet](https://developers.google.com/analytics/devguides/collection/analyticsjs/advanced#alternativeSnippet) that's listed on the developer documentation site. Since it's not exactly the same, I'll go through the differences. If H5BP chooses to promote a slightly different snippet, I think it would make sense to explain the differences and the trade-offs. I'd be willing to submit a separate PR for that.

#### Differences from GA's async snippet:

1. Adds the `defer` attribute.
2. Uses an undeclared variable assignment rather than assigning the `ga` function on the `window` object.
3. Removes the check to see if `ga` already exists.
4. Removes the newline between the `ga('create', ...)` and `ga('send', ...)` invocations.
5. Uses "https" rather than a protocol-less script source.

#### Thoughts from the GA team:

1. The `defer` attribute is tricky. The advantage of defer is it adds non-blocking support for older browsers. The disadvantage is it changes the execution time of the analytics.js script on *all* browsers. Deferred scripts will execute after the entire document has been parsed and all other (non-deferred) scripts have been downloaded and executed.

  The problem with this, from an analytics perspective, is it will affect pageview counts and data quality, especially on sites that load a lot of JavaScript. If a user comes to your site and leaves before all the JavaScript has finished downloading, that visit will be lost. This may not be a big deal to some analytics users, but it is a big deal to many, and that's why it's not in the official async snippet.

2. Removing the `window` prefix from the initial assignment is definitely not a good idea. If a site has the `'use strict'` flag anywhere in the global scope, this code will throw a reference error. You can see that happening here: http://jsbin.com/fohexifaku/1/edit?html,console

  I agree with @mathiasbynens about it being overkill to use `window.ga` everywhere, but to avoid the possibility of a reference error, it would be better to keep it, at least for the assignment expression.

3. A lot of people who use the analytics.js snippet copy and paste the code carelessly, use plugins or widgets that add their own analytics.js tracking to the page, or do other things that result in the `ga` function already being defined by the time the snippet code is executed. To avoid overriding existing code, I think we need to keep the check in place.

4. This is really minor, but one of the challenges we're trying to overcome is how to educate developers that it's possible to do so much more with Google Analytics than what is provided out of the box by the snippet. IMHO, anything that's done to obfuscate these two lines of code will further contribute to perpetuating this misunderstanding. Again, I know this is minor.

5. Using https is a good idea. Very soon http://www.google-analytics.com/analytics.js will permanently redirect to https, and at that point we'll change all our snippets to be https-only. I see no reason H5BP shouldn't do this now though.

To wrap this up, here's what I'd recommend H5BP use as its default analytics.js snippet:

```html
<script async src='https://www.google-analytics.com/analytics.js'></script>
<script>
window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
ga('create', 'UA-XXXXX-Y', 'auto');
ga('send', 'pageview');
</script>
```

Again, if the decision is made to add `defer` to the script tag, I think a note about how it may affect the quality of analytics data would be warranted.

mathiasbynens:
@philipwalton 

> 2. Removing the `window` prefix from the initial assignment is definitely not a good idea. If a site has the `'use strict'` flag anywhere in the global scope, this code will throw a reference error. You can see that happening here: http://jsbin.com/fohexifaku/1/edit?html,console

“Anywhere in the global scope” is not correct; that’s now how it works. See e.g. <http://jsbin.com/sodenokeki/1/edit?html,console>. There is no possibility of a reference error unless the user explicitly self-sabotages and adds `'use strict'` to the `<script>` block containing the GA snippet.

philipwalton:
> I often found annoying that GA triggers the gif request at DOMContentLoaded which delays the window.onload event for a minimum latency of 100ms, up to a second.

@hexalys no request is sent at `DOMContentLoaded`. The only request sent to Google Analytics is the one made by the method `ga('send', 'pageview')`, which happens as soon as analytics.js is downloaded and executed.

There are cases where the gif request will delay the `load` event, but those can be avoided by setting the [transport](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#transport) option to `'beacon'`, which will attempt to send the hit with the [`navigator.sendBeacon`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) method, if the browser supports it.

The `sendBeacon` method will likely become the default method used by analytics.js in the near future, so if you simply do nothing and keep using the recommended snippet, this problem will simply go away.

I've added some screenshots so you can see the difference:

```js
ga('send', 'pageview');
```

screen shot 2015-04-24 at 6 50 22 pm

```js
ga('send' ,'pageview', {transport: 'beacon'});
```

screen shot 2015-04-24 at 6 50 34 pm

Notice how in the second screenshot the `load` event happens before the hit is finished being sent.

philipwalton:
> “Anywhere in the global scope” is not correct; that’s now how it works. See e.g. http://jsbin.com/sodenokeki/1/edit?html,console. There is no possibility of a reference error unless the user explicitly self-sabotages and adds 'use strict' to the <script> block containing the GA snippet.

@mathiasbynens ahh, I stand corrected.

Though I suppose if a user decides to copy that part of the snippet into their main JavaScript file and include it somewhere that's already in strict mode, you'd have the same problem.

It still seems safer to me to keep the window prefix. Do you disagree?

mathiasbynens:
@philipwalton

> Though I suppose if a user decides to copy that part of the snippet into their main JavaScript file and include it somewhere that's already in strict mode, you'd have the same problem.

True. IMHO this seems unlikely though; global `'use strict'` is a bad practice, and if someone wants to move the GA snippet to a `.js` file, they’d probably want to move the script loading (`<script async …>`) out of the HTML as well, as the old script does.

> It still seems safer to me to keep the `window` prefix in the assignment. Do you disagree?

Agreed. I’ve updated the PR.

hexalys:
@philipwalton
> no request is sent at DOMContentLoaded.

Ok, indeed sorry. I was under the mistaken assumption it had to wait for DOMContentLoaded to record the event for PageSpeed, and also because I previously had it triggered that way.

>There are cases where the gif request will delay the load event, but those can be avoided by setting the transport option to 'beacon'.

I didn't catch that. Thanks. However, it only solves the issue in half the browsers in the short-to-medium term. Given that my ga logic in an external scripts, unpredictable timings are not easily manageable anyway, so I may stick to my approach. With simple fast loading sites, I am not very inclined to track visits that have not reached window.load.

> The advantage of defer is it adds non-blocking support for older browsers. The disadvantage is it changes the execution time of the analytics.js script on all browsers.

When `defer` is at the top yes. But not the way h5bp is now. Deferred scripts execute right after the last script before `</body>`, so any possible blocking is over at this point. The way they have `defer` at the bottom will execute at the same time as without `defer`. Unless delayed by other prior/above defer scripts, a `defer` at the bottom isn't [technically making any difference](https://github.com/h5bp/lazyweb-requests/issues/42#issuecomment-1779556), and it must execute right away. `defer` becomes part of the "list of scripts that will execute when the document has finished parsing".

ramasilveyra:
In the PR why the GA snippet from dist/index.html and src/index.html are differents?

alrra:
> In the PR, why the GA snippet from dist/index.html and src/index.html are differents? One has window.ga and the other simply ga

@ramasilveyra Fixed, thanks! (@mathiasbynens just forgot to run the build step after he [updated this pull request](https://github.com/h5bp/html5-boilerplate/pull/1696#issuecomment-96121344), but as long as the content in `src/` is correct, once something is merged, @h5bp-bot will automatically update the `dist/` directory to reflect the changes from `src/`).

PR #1709, opened 2015-05-29
Update plugins.js

ramtml:
i have updated plugin design inspired from Addy osmani's smashing magazine article. 

http://www.smashingmagazine.com/2011/10/11/essential-jquery-plugin-patterns/

### DIFF
--- a/src/js/plugins.js
+++ b/src/js/plugins.js
@@ -1,24 +1,38 @@
 // Avoid `console` errors in browsers that lack a console.
-(function() {
-    var method;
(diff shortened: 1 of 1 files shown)


drublic:
Thanks. This replaces the console stub we've included in plugins.js. Apart from that it adds the jQuery plugin pattern which we don't want to support here.
Please add your own `jquery.plugin.js` file if you want to write a jQuery plugin.

So far we are thankful for the contribution but are not willing to include it. Thank you!

PR #1715, opened 2015-06-22
::first-letter and ::first-line pseudo-elements are now printed in black

SidRoberts:


drublic:
Thanks for the contribution. Makes sense to support these pseudo elements too.

Currently we still support IE8 and thus have to use `:first-letter` and `:first-line` I think.
Apart from that, could you please add a note in our `CHANGELOG.md` file?

SidRoberts:
@drublic Done. :smile:

drublic:
Thank you, @SidRoberts! Any other opinions?

SidRoberts:
Nope, I'm happy. :smile:

drublic:
Thanks again for your work. Now finally landed in master.

PR #1720, opened 2015-07-04
Doc: Minor Update

JamesIves:
I've updated the html.md document with additional data on the title and meta description tag. This entire document details every other element besides these two so I felt it was appropriate to include them. I also adjusted some of the formatting on the headers to make them more consistent. If this type of change is welcomed I'll gladly comb through the rest of the documents and attempt to improve them where I can.

### DIFF
--- a/dist/doc/html.md
+++ b/dist/doc/html.md
@@ -43,7 +43,7 @@ will be displayed within the browser and in search engine results.
 
 ### The order of the `<title>` and `<meta>` tags
 
-The order in which the `<title> and the `<meta>` tags are specified is
+The order in which the `<title>` and the `<meta>` tags are specified is
 important because:
 
 1) the charset declaration (`<meta charset="utf-8">`):
--- a/dist/doc/html.md
+++ b/dist/doc/html.md
@@ -13,7 +13,7 @@ By default, HTML5 Boilerplate provides two `html` pages:
 ## `index.html`
 
 
-### The `no-js` class
+### The `no-js` Class
 
 The `no-js` class is provided in order to allow you to more easily and
 explicitly add custom styles based on whether JavaScript is disabled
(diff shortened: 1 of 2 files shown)


clarkdemvher:
Sent from Samsung tablet.


-------- Original message --------

src/doc/html.md
```diff
@@ -30,9 +30,20 @@ attribute to `<html>` as in this example:
 <html class="no-js" lang="en">
 ```
 
+
+## Document Title
```
arthurvr: This makes me wonder how far our documentation should go and at what point it starts to look like an intro to HTML. This looks like general knowledge which would better fit in a getting started with web development tutorial then in the H5BP docs.

On the other hand, it's important to emphasise  things like this. What does the rest think?

clarkdemvher: Sent from Samsung tablet.


-------- Original message --------

dist/doc/html.md
```diff
@@ -30,9 +30,20 @@ attribute to `<html>` as in this example:
...
+
+```html
+<title>HTML5 Boilerplate</title>
+```
+
+
 ### The order of the `<title>` and `<meta>` tags
 
-The order in which the `<title>` and the `<meta>` tags are specified is
+The order in which the `<title> and the `<meta>` tags are specified is
```
drublic: misses closing `.

src/doc/html.md
```diff
@@ -30,9 +30,20 @@ attribute to `<html>` as in this example:
...
+
+```html
+<title>HTML5 Boilerplate</title>
+```
+
+
 ### The order of the `<title>` and `<meta>` tags
 
-The order in which the `<title>` and the `<meta>` tags are specified is
+The order in which the `<title> and the `<meta>` tags are specified is
```
drublic: Ah yea, see above.

src/doc/html.md
drublic: I think it's a good addition. Mostly people will know it but maybe forget about the search engine results.

src/doc/html.md
JamesIves: Fixed it up.

alrra:
Thanks @JamesIves! :purple_heart:

PR #1722, opened 2015-07-14
Trims comment lines to 79 characters in main.css

jfinkels:
This prevents annoying line continuations on default 80 character width terminals when using emacs (for example).


robbbz:
> 79

If your limit is 80, 80 should work fine, no?

jfinkels:
Not in emacs, maybe it works in other text editors.

robbbz:
Your problem, then. H5BP shouldn't need to care about this if it's specific to one editor.

drublic:
Thanks for reporting.
I did a quick search on this issue and didn't find anything regarding the line length bug. Do you have a link for me describing the behavior?
I think this is something which should be solved by the editor though.

Thank you anyways.

jfinkels:
@drublic It's not a bug, it's a feature of emacs with its default configuration when opened in a standard 80 character terminal. Here is the description of [line continuation in emacs][1], here is a [stackoverflow issue][2] describing the behavior. I understand that this is an editor issue, but thought it was a quick fix to a comment line that prevents weird behavior when using the default configuration of a somewhat common text editor. Thanks for taking a look anyway.

[1]: http://www.gnu.org/savannah-checkouts/gnu/emacs/manual/html_node/emacs/Continuation-Lines.html
[2]: http://stackoverflow.com/questions/7153874/stop-emacs-from-wrapping-80-column-lines-in-an-80-column-terminal

PR #1729, opened 2015-08-08
add keywords in package.json

PeterDaveHello:


kevva:
This isn't published to npm so it doesn't makes sense.

PeterDaveHello:
It won't bite :smile:

kevva:
No, but it's simply not useful.

PeterDaveHello:
In fact, I saw a request to add this framework on public cdn, it'll be nice to have keywords in package.json!

drublic:
HTML5 Boilerplate is more of a template than a framework. What would you
add to a CDN or npm registry?
I don't think we need this.

PeterDaveHello:
Ok, thanks!

PR #1734, opened 2015-09-10
Minor comment tidy up

michaelthorne:


drublic:
Hi. Thank you for the suggestions. I think we're good with the current comment style since it's consistent within the project.

michaelthorne:
@drublic no worries, but:

The semi colon at the end of the comment for the `.hidden` class was left over from [this commit](https://github.com/h5bp/html5-boilerplate/commit/422537b96c61d1e37371becc97769d9df592b4ac) where the URL was removed.

And those were the only two single line, block style comments with fullstops at the end. None of the others in `main.css` have at the end of single line comments.

:+1:

alrra:
> The semi colon at the end of the comment for the .hidden class was left over from this commit where the URL was removed.

Thanks @michaelthorne. Fixed in https://github.com/h5bp/html5-boilerplate/commit/1cf48987f5932e730e77f16a85e4b9235f97809f.

Issue #1547, opened 2014-06-11
Why Google analytics and not, for example, Piwik?


ghost:
This is an honest question, because I'm wondering what the reasons are for HTML5bp's use of Google Analytics. I've searched for it and haven't found any issues referencing this, or any specific documentation on it, which is why I've opened this issue.

I personally think that Piwik's interface is more user friendly and easier to understand. It isn't difficult to use or set up at all. And most importantly, it is completely open source and more respectful of visitors' online privacy (i.e. data isn't automatically shared with a third party). It is a _little_ easier to implement tracking on ones website using GA, but I don't think that that warrants a complete omission of the alternatives.

I think it would be a good idea (and more transparent) to either
- explain why Google analytics is chosen over other analytics software
- list the alternatives (if there is no specific reason)
- or switch to an analytics script that is more in line with open source philosophy

Issue #1275, opened 2012-12-21
Remove Google Analytics


KaptajnKold:
It concerns me that H5BP promotes the use of Google Analytics as a best practice by including the GA snippet in the boilerplate. There a privacy issues to be considered when using GA and I for one am not comfortable with the fact that more than 50% of all websites are sending information about their visitors to the same third party. The ubiquity of GA gives Google the ability to track individuals browsing from site to site and while the Analytics tools themselves are nifty for web masters wanting to optimize their sites, they come at a privacy cost for their users.

In short, I think the snippet should be removed lest we want web masters to mindlessly infringe on their users privacy because they think it's somehow best practice. Let it be up to the web masters to consciously choose which tracking tools if any they want to employ.

guilhermesimoes:
Funny how just 6 months after this issue was closed we got the Snowden revelations. I think @KaptajnKold's arguments are even stronger today than they were 3 years ago.

> you haven't provided a strong argument about the privacy risks either.

Is this enough? [NSA uses Google cookies to pinpoint targets for hacking, the Washington Post reports](http://www.washingtonpost.com/blogs/the-switch/wp/2013/12/10/nsa-uses-google-cookies-to-pinpoint-targets-for-hacking/).

I think this issue could be reopened so we could have an honest discussion about the implications of using Google Analytics.

Volker-E:
I agree strongly with @KaptajnKold argument on the popularity of HTML5 Boilerplate -- that it makes beginners think what's the standard/it sets the standards without them necessarily asking for the why. 
Why explicitly integrating Google Analytics and not Piwik code for your own server? 
For me that should be more a side-fork like different server configurations of H5BP.

roblarsen:
>Why explicitly integrating Google Analytics and not Piwik code for your own server? 

*Potential privacy issues aside*, the two technologies aren't even in the same conversation when it comes to usage. Offering GA as part of the base HTML5  Boilerplate package serves a majority- including many millions of people who don't have access to their own server to install their own analytics package. The HTML5 Boilerplate project selling Piwik to people is just a crazy idea especially since many (most?) developers aren't the ones actually making the decision about what analytics package to use,

http://trends.builtwith.com/analytics/Google-Analytics > 50% of the web

http://trends.builtwith.com/analytics/Piwik-Web-Analytics > 0.5% of the web

Volker-E:
[...]

PR #1750, opened 2015-11-03
Add Google Tag Manager instead of Google Analytics

JohnnyWalkerDesign:
Google Tag Manager allows for better control over inserted tags, like Google Analytics, as well the ability to quickly add/update/remove tags without needing to modify your code. Preferable?

From Google's documentation:

*“Google Tag Manager is an asynchronous tag, meaning that when it executes, it does not block other elements from rendering on the page. It also causes the other tags that are deployed via Google Tag Manager to be deployed asynchronously, meaning that a slow loading tag won’t block other tracking tags.”*

https://developers.google.com/tag-manager/devguide

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -27,14 +27,13 @@
         <script src="js/plugins.js"></script>
         <script src="js/main.js"></script>
 
-        <!-- Google Analytics: change UA-XXXXX-X to be your site's ID. -->
-        <script>
-            (function(b,o,i,l,e,r){b.GoogleAnalyticsObject=l;b[l]||(b[l]=
-            function(){(b[l].q=b[l].q||[]).push(arguments)});b[l].l=+new Date;
-            e=o.createElement(i);r=o.getElementsByTagName(i)[0];
-            e.src='//www.google-analytics.com/analytics.js';
-            r.parentNode.insertBefore(e,r)}(window,document,'script','ga'));
-            ga('create','UA-XXXXX-X','auto');ga('send','pageview');
-        </script>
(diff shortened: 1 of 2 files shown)


QWp6t:
:-1:

JohnnyWalkerDesign:
Thanks for the well-reasoned and considered addition to this discussion. I was a skeptic when I started reading, but you convinced me by the end.

webhat:
Implementing Google Tag Manager is none trivial, and in my experience there is no great benefit if you are not running a site that requires all the enhancements such as an ecommerce site.

lkraav:
Indeed. Tag management is for people who know they need it, not so much suitable for a hardcoded default thing. Raising awareness, always a good thing, could still be accomplished by a documentation patch.

JohnnyWalkerDesign:
I wouldn't say that GTM is "non trivial". It's slightly more complicated than GA, but only slightly, and is a solution that's future proofed (as far as we can tell) once done. Still, if the general consensus is that it's not an improvement, fair enough.

roblarsen:
FWIW, there have been a couple of previous discussions of removing Google Analytics altogether or replacing it with other opens #1275 #1547 #1548

JohnnyWalkerDesign:
Well this isn't about removing GA, and wouldn't appease those who aren't fans of Google :)

roblarsen:
I'm just adding links to relevant issues from the past.

paulirish:
Thanks for the PR. 

I've actually never used GTM, personally. Swapping these snippets is pretty easy, so I think it's alright to leave our optimized GA snippet in place for the default user.

PR #1755, opened 2015-11-17
Minor fix on closure syntax

blckshrk:


drublic:
Thanks for the suggestion. It think this is a personal preference here.
I personally think `(function () {}())` is easier to understand than `)(function () {})()` since it reminds more of `functionCall()`. Anyway I don't really care about what is used though.
In this case I think we can leave it the way it is.

Thanks again for taking the time for the PR.

FagnerMartinsBrack:
For additional information, check https://youtu.be/eGArABpLy0k?t=1m10s

QWp6t:
wo

@drublic since when are you owner of this repo? congrats. :tada: :balloon:

drublic:
@QWp6t Thank you :) <3

blckshrk:
@drublic You're right, it's only a matter of personal preferences. I found the second style more readable, but that's my opinion.

@FagnerMartinsBrack Thanks for the reference ;-)

PR #1788, opened 2016-02-11
use es2015 syntax in mocha tests

marcobiedermann:
use `babel` to use es2015 syntax in mocha tests
- use `import` instead of `require`
- use fat arrow `=>` function instead of `function() {}` callback
- use and `let` and `const` instead of `var`
- use template stings instead of string concatenation


chrislaughlin:
:+1:

paulirish:
Thanks for writing the PR. 

I'd be interested in accepting this if its not dependent on babel. In my experience adding transpilation into your test runner makes debugging fairly painful.

If you want to confirm with node.green that all this stuff works in stable node 4.4 then let's just rip out the babel bit and ship it.

marcobiedermann:
@paulirish Thanks Paul for your comment.
As far as I know there is no other way than using a transpiler like Babel to use `import` state in JavaScript.
All other features: `arrow functions`, `let`, `const` and `template strings` are supported in Node 4.4+
The only solution dropping Babel would be to stick with `require` instead of `import`.

Do you know another way to get around this issue?

marcobiedermann:
@paulirish
I've merged the current master into my PR to keep it up to date.
I've also noticed that you already include Babel.js so we are good with using `import` in tests.

PR #1793, opened 2016-02-25
Remove obsolete meta tag

camskene:
This PR removes the meta tag for `x-ua-compatible`.

__Why?__

This tag was introduced with IE8 for 'Compatibility View' - a mechanism for the browser to choose a rendering mode.

With MS [discontinuing support](https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support) for older browsers, perhaps this tag is no longer necessary?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,7 +2,6 @@
 <html class="no-js" lang="">
     <head>
         <meta charset="utf-8">
-        <meta http-equiv="x-ua-compatible" content="ie=edge">
         <title></title>
         <meta name="description" content="">
         <meta name="viewport" content="width=device-width, initial-scale=1">


albfernandez:
I think it's good at this place. It has no effect in non-IE browsers, and force all IE from 8 to 11 to use de most recent rendering engine.
It can be a problem in intranet sites where IE8-11 uses "compability mode" (IE 7) by default. With this tag you force IE to use most recent rendering engine, wich made easy cross-browser development.

paulirish:
I'm interested in removing it. I know that we added it because IE would kick sites into compat mode even with a doctype. We need to research to find out why that happened and if it happens in IE8+.

coliff:
This meta tag should stay in the project for now.

Rather than regurgitate the reasons here - there's a great Stack Overflow post about it:
http://stackoverflow.com/questions/6771258/what-does-meta-http-equiv-x-ua-compatible-content-ie-edge-do

roblarsen:
I'm going to close this. The Stack Overflow post mentioned above is enough evidence- especially since IE8 support isn't out of the project yet. It does no harm except in absence. 

FWIW, I've also got anecdotal evidence. A project I'm on _just_ had to add this meta tag.  Thankfully it wasn't me that had to add it or I would have had "building intranet applications for banks" flashbacks.

PR #1798, opened 2016-03-02
IE jQuery

ZakariaRidouh:


coliff:
Hi that code suggestion would still load jQuery v2 for IE8 users though. To only load jQuery v1 for IE8 and below and jQuery v2 to all other browsers you need to use conditional statements like this:
https://gist.github.com/coliff/b80669f114b8ccb8560e

PR #1799, opened 2016-03-02
Fix print styles for Internet Explorer 11

damianwajer:
*:first-line pseudo-selector freezes window with print preview in IE 11 (tested on Windows 7 with IE 11.0.9600.17420 and 11.0.9600.18204, Windows 8.1 with IE 11.0.9600.16663, Windows 10 with IE 11.63.10586.0)

To reproduce just open: http://demo.html5boilerplate.com/ or /dist/index.html and select Print > Print preview...

### DIFF
--- a/dist/css/main.css
+++ b/dist/css/main.css
@@ -213,7 +213,10 @@ textarea {
     *:before,
     *:after,
     *:first-letter,
-    *:first-line {
+    p:first-line,
+    div:first-line,
+    blockquote:first-line,
+    li:first-line {
         background: transparent !important;
         color: #000 !important; /* Black prints faster:
                                    http://www.sanbeiji.com/archives/953 */
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -205,7 +205,10 @@ textarea {
     *:before,
     *:after,
     *:first-letter,
-    *:first-line {
+    p:first-line,
+    div:first-line,
+    blockquote:first-line,
+    li:first-line {
         background: transparent !important;
         color: #000 !important; /* Black prints faster:
                                    http://www.sanbeiji.com/archives/953 */


slavanga:
The freeze is caused by the star selector. Scoping the :first-line and :first-letter selectors to paragraphs seems to be good compromise.

```css
*,
*:before,
*:after,
p:first-line,
p:first-letter {
	background: transparent !important;
	color: #000 !important;
	box-shadow: none !important;
	text-shadow: none !important;
}
```

damianwajer:
@slavanga Good point. From what I've tested it seems that only `*:first-line` causes this bug so maybe just `p:first-line`?

```css
*,
*:before,
*:after,
*:first-letter,
p:first-line {
    background: transparent !important;
    color: #000 !important; /* Black prints faster:
                               http://www.sanbeiji.com/archives/953 */
    box-shadow: none !important;
    text-shadow: none !important;
}
```

Any other opinions?

paulirish:
Can someone look up the history of why those first-* lines are in here? Rather unexpected, IMO.

slavanga:
I found the original pull request:

roblarsen:
Thanks @slavanga 

I was able to reproduce this . Annoying issue. It works in Edge. 

@damianwajer I think expanding from just p:first-child to include maybe `div`, `blockquote` and `li` would cover a good majority of the intent behind the universal selector.  Want to (months alter, sorry!) alter the PR to include those other common elements?

damianwajer:
@roblarsen 

>  I think expanding from just p:first-child to include maybe `div`, `blockquote` and `li` (...)

Did you mean `p:first-line`? 

```css
*,
*:before,
*:after,
*:first-letter,
p:first-line,
div:first-line,
blockquote:first-line,
li:first-line {}
```
Like this? It works in IE 11.

roblarsen:
@damianwajer Yeah, that's it.  I don't want to add a hundred elements there as it's just a print style and the point of the PR is to fix the freeze. I think those elements are going to cover a majority of the places where people use a `:first-line`

damianwajer:
@roblarsen I see, done.

roblarsen:
Thanks!

PR #1806, opened 2016-03-09
Change "adaptable" to "adaptive" in README.MD

alex-teren:


paulirish:
I don't think there's strong consensus either way. Gonna keep as is, since this term is used in a few other spots, as well.

PR #1814, opened 2016-03-29
Change license format to SPDX format

wolfie:
The old license style (`license.type` and `license.url`) is [deprecated by NPM](https://docs.npmjs.com/files/package.json#license). Using [SPDX format](https://spdx.org/licenses/MIT.html) instead


paulirish:
thx

PR #1828, opened 2016-04-17
Add web application manifest file (WIP)

ghost:


paulirish:
manifest.json is the typical filename. if someone wants to pick this PR up with those changes we could discuss it further.

PR #1834, opened 2016-05-03
Include Segment Snippet alongside GA snippet

isisAnchalee:
This adds the [Segment](https://segment.com/docs/libraries/analytics.js) snippet alongside the Google Analytics snippet. This is better because Segments allows you to send data to multiple analytics tools, including Google Analytics, without having to add any other scripts to the page.


jpdevries:
I've noticed Google Analytics punishes your Google PageSpeed score because they don't cache the script long enough. Rather ironic.

paulirish:
neat, but segment is too nice for us here.

thanks for writing the PR!

PR #1845, opened 2016-05-19
Remove sudo: false in Travis configuration

kaelig:
Travis builds now run on their container by default, meaning `sudo: false` is now the default config value.


paulirish:
neat. sgtm

PR #1856, opened 2016-05-30
Added theme-color meta tags and OG meta tags

kalpeshsingh:
Hello devs,

I'm very excited 😍 as this is my first pull request on GitHub. I have added two fancy yet needed snippet in my pull request.

Let's discuss - 

#### 1. `theme-color` meta tag

According to Google Developer site - 

> Chrome, Firefox OS, Safari, Internet Explorer and Opera Coast allow you to define colors for elements of the browser

So I thought, why not we (h5bp) should provide this to our users. I know this is a fancy feature, but we all are here to provide latest and best.

##### Official docs 

[Google Developers](https://developers.google.com/web/fundamentals/design-and-ui/browser-customization/theme-color?hl=en)

#### 1. `og` meta tag

According to Open Graph Protocol site - 

> The Open Graph protocol enables any web page to become a rich object in a social graph.

When I downloaded h5bp at that time I presumed that I will get a social-friendly blueprint, and frankly I disappointed a bit. We should provide our users a social-friendly structure. So, I decided to add OG meta tags.

The benefits of these tags are that you can control how your website or app will look and whether it shows contents that convey your message when someone shares on social media platform such as Facebook, Twitter etc. The best part is that many social media follow OG Protocol and in case if you miss some specific site's sharing guideline then they can make use of OG meta tags.

##### Official docs 

[Open Graph Protocol](http://ogp.me/)

Excited to discuss! 😊

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,8 @@ Choose one of the following options:
   errors in older browsers.
 * "Delete-key friendly." Easy to strip out parts you don't need.
 * Extensive inline and accompanying documentation.
-
+* Open Graph Meta tags. Now boost your social media presence. [NEW]
+* Color metas make mobile browser toolbar to match your website color scheme. [NEW]
 
 ## Browser support
 
--- a/src/index.html
+++ b/src/index.html
@@ -16,6 +16,12 @@
         <!-- iOS Safari -->
         <meta name="apple-mobile-web-app-status-bar-style" content="#101524">
         
+        <!-- Open Graph Protocol - Decide how your page should display on Facebook and other social media -->
+        <meta property="og:title" content="your jaw-dropping title here" />
+        <meta property="og:type" content="website" />
+        <meta property="og:url" content="you can either use your meta or use sexy and entertaining copy here" />
+        <meta property="og:image" content="a high quality image here" />
+        
         <link rel="apple-touch-icon" href="apple-touch-icon.png">
         <!-- Place favicon.ico in the root directory -->
         
--- a/src/index.html
+++ b/src/index.html
@@ -6,10 +6,19 @@
         <title></title>
         <meta name="description" content="">
         <meta name="viewport" content="width=device-width, initial-scale=1">
-
+        
+        <!-- theme-color meta tag to set the toolbar color in mobile broswer -->
+        
+        <!-- Chrome, Firefox OS, Opera and Vivaldi -->
+        <meta name="theme-color" content="#101524">
+        <!-- Windows Phone -->
+        <meta name="msapplication-navbutton-color" content="#101524">
+        <!-- iOS Safari -->
+        <meta name="apple-mobile-web-app-status-bar-style" content="#101524">
+        
         <link rel="apple-touch-icon" href="apple-touch-icon.png">
         <!-- Place favicon.ico in the root directory -->
-
+        
         <link rel="stylesheet" href="css/normalize.css">
         <link rel="stylesheet" href="css/main.css">
         <script src="js/vendor/modernizr-2.8.3.min.js"></script>


quantumpacket:
Hi, @kalpeshsingh

OG tags are already specified here: https://github.com/h5bp/html5-boilerplate/blob/master/dist/doc/extend.md#social-networks

Theming is already specified here: https://github.com/h5bp/html5-boilerplate/blob/master/dist/doc/extend.md#theme-color

I would recommend that you expand on those docs instead of adding them to the HTML templates as defaults since their use is subjective.

kalpeshsingh:
Hey @quantumpacket ,

I believe that OG tags shouldn't be subjective. The users should have a HTML structure that is W3C valid, developer friendly as well SEO friendly too.

If we put OG tags in the extended version of boilerplate then users mayn't add to their boilerplate.

Here, I *strongly* advocate of putting at least OG tags in main HTML templates.

quantumpacket:
> Here, I strongly advocate of putting at least OG tags in main HTML templates.

OG tags are used for social media sites, and is not a W3C spec as far as I am aware. It was made for Facebook and they are the primary user of it. For SEO purposes you should be using JSON-LD or Schema.org tags. Not every site using this boilerplate is going to need social media tags, that is probably the best reason to not add it by default.

kalpeshsingh:
I'm afraid that you misunderstood me @quantumpacket .I said we should contribute towards SEO friendly structure. The current structure is W3C valid and developer friendly, but not SEO friendly.

> OG tags are used for social media sites, and is not a W3C spec as far as I am aware.

I strongly disagree that OG tags are used for social media sites. OG tags are normally used to control over title, image, description etc. Almost all user wants their site to be displayed properly on social giants like Facebook, Twitter etc.

> It was made for Facebook and they are the primary user of it. 

**It was made by Facebook for Facebook and other social media sites,** not merely for Facebook. Please understand that it is open protocol and anybody (Twitter, Pinterest etc.) can use this protocol and they do use.

**So, everyone wants their site to be displayed properly on Social Media and almost all boilerplate users need it. This is my reason to put OG tags in the main template.**

quantumpacket:
@kalpeshsingh  Did you even bother to check out the link in my last post? This has been brought up before and it was rejected. I'm not going to waste anymore time arguing on this subject. You're trying to justify this based on *"your needs"*, and not the needs of the 34k+ users of this boilerplate. The boilerplate should be generic to best fit everyone's use cases, and if they choose to do so they can read the extended docs and add upon it. This is how it is currently setup. There are plenty of sites that do not require social media support.

rsp:
@kalpeshsingh, that's not the best reaction for a first pull request... ;) Don't worry, it happens. :)

Of course not everyone needs social media tags, like not everyone needs the Apple touch icon, jQuery, Modernizr and Google Analytics - and yet they are all present by default, presumably because it's easier to remove them when not needed than to remember what to put there if you need them.

Hell, I remember when a hot pink selection background color was used by default :) so the defaults here have always been somewhat... controversial. ;)

As for me, I think it's easier to remove those 4 lines if I don't need them than to search for the documentation on what should I put where (if I not forget about it at all). I remember that the case for a default selection color was that it's easier to remove it or change it than to remember what to add and where. The same goes for Google Analytics - if I don't need it I just remove it and don't complain. I would treat the social media tags the same way.

I would wait for h5bp contributors to say what they think about your PR but just in case it gets rejected then please don't remove your fork because I'd like to have it all in one place whenever I need to build a website that I want to work nicely with social media - which is to say, most of the time.

An ideal solution for me would be to have a `social.html` file along with `index.html` for people to choose but if that idea is rejected then maybe you should maintain a fork called `social-h5bp` (or something like that) which would be synced with `h5bp` but with all social media tags and features possible? Just a thought. I would use it myself. Mention me in a comment if you make a new repo - I already watch your https://github.com/kalpeshsingh/html5-boilerplate

Best regards and don't worry about the reaction to your first PR. ;) It can only get better now. :) Good luck.

kalpeshsingh:
@rsp haha! Indeed not best reaction! I have written *Excited to discuss* but the discussion turned to be exploded.  

> ...it's easier to remove them when not needed than to remember what to put there if you need them.

This is what I tried to address. 

Even I'm waiting for official response for this PR and in case it gets rejected I will keep forked version and keep it updated with only relevant stuff. 

Thanks for writing such long energetic reply. I'm fueled with energy now. ;)

*(Pssst... Developers are lazy then how can you manage to write such long motivational reply :tongue: )*

QWp6t:
This has been discussed a number of times. I'm fairly certain the consensus right now is: add it to the docs.

I can almost guarantee this won't be accepted, especially considering it's effectively limited to just Facebook. I think there's much greater use/demand for SEO stuff like Schema.org (a collaborative effort by the 3 largest USA search engines) than for OG (Facebook only).

kalpeshsingh:
Hi @QWp6t ,

> ...it's effectively limited to just Facebook. 

Please understand that it is made by Facebook and can be used by others. These tags are protocols and open sourced. 

From my previous comment -


> It was made by Facebook for Facebook and other social media sites, not merely for Facebook. Please understand that it is open protocol and anybody (Twitter, Pinterest etc.) can use this protocol and they do use.

paulirish:
hey kalpesh. I agree with the other folks here that this is best kept to the docs.

Early h5bp was "delete-key friendly" but it's evolved into a more svelte setup by default, with docs that provide more assistance when useful. For example I'd love to read docs that explain the difference between the OG title and the `<title>` tag and when the former is used. That's going to be best described where you can link to supporting docs, so yeah.. markdown ftw.

Also, old projects have a lot of history. Despite you being a new contributor its hard for folks to be nice when time is so limited. It happens. Don't take it personally. :)

kalpeshsingh:
Thanks Paul! 
No hard feelings and respect to all contributors! :smile: 
I will keep my eyes on updates and will come up with new PR with concrete base (if any).

PR #1863, opened 2016-06-09
Add common large file format pattern matches to .gitattributes

davewallace:
Specifies patterns for matching files which can be stored using Git
Large File Storage to reduce workflow overhead. The default files are
binary, but this is not mandatory for any given file.


paulirish:
I like it but since so few folks are using git LFS i don't think we're ready to have this in here yet.

PR #1876, opened 2016-07-16
Reduce statements, improve styling in plugins.js.

kmcconnell:
Reduces the number of statements by removing superfluous assignments to vars that are only used once. Removed gratuitous parentheses as well as obsolete console methods to reduce iterations through the array, and placed elements in alphabetically where out of place.

### DIFF
--- a/src/js/plugins.js
+++ b/src/js/plugins.js
@@ -1,24 +1,17 @@
 // Avoid `console` errors in browsers that lack a console.
-(function() {
-    var method;
-    var noop = function () {};
+(function () {
     var methods = [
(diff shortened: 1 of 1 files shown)


paulirish:
h5bp supports ie8 currently which doesn't have forEach.

Also looks like you're creating an implicit global of `console` since you left out the `var`

if you want to update to fix those, we can merge this.

paulirish:
thanks again for the PR :)

roblarsen:
@kmcconnell we're removing IE8 support, so all we're missing to pull this in the `var` as mentioned above. 

Thanks!

PR #1878, opened 2016-07-21
Restructure gulp and more features for best apps

seokirill:
Restructure gulp: tasks in foreign task-files.

CSS & JS on create bundles, minify&optimize if use
`$ gulp production`

Webpack 2 and Babel, csso and postCss make your apps better.

And all it works at start!

Be better easely


les2:
Why is the file mode changed for several markdown files (adding execute bit)?

seokirill:
Dunno (:
Some strange things was my machine.
It isnt Critical

paulirish:
This is a fairly massive change. If you're serious then we'd need a issue posted up front to discuss what you'd like to change/improve.

That said, I'm nearly certain it wouldn't work out. Web Starter Kit is more similar to what you're driving for here.  H5BP will stay more barebones.

Issue #1779, opened 2016-01-21
Remove crossdomain.xml


frontenddeveloping:
Browsers disabled flash by reason of it secuire problems (but enabled it after fixes). Note sure that it still nessesary.
Let's push browser native capabilities, not plugins.

paulirish:
sure. sounds good. it's probably time.

anyone wanna file a PR for this?

roblarsen:
closed via #1881

BobbyBabes:
`crossdomain.xml` is still in the v5.3.0 download here at GitHub in `Releases`, and under the orange download button at http://html5boilerplate.com

roblarsen:
We haven't done a release, unfortunately.

BobbyBabes:
OK, that explains it then.

PR #1881, opened 2016-08-08
Delete crossdomain.xml

t1st3:
from #1779


roblarsen:
@t1st3 Can you squash these commits?

t1st3:
I just squashed them @roblarsen

roblarsen:
Thanks!

PR #1882, opened 2016-08-09
Remove print img max-width: 100%. 

### LINKED ISSUE #1741: Print media query for img max-width breaks Google Maps printing in Chrome
Somewhat related to the issue with images in IE8 (https://github.com/h5bp/html5-boilerplate/issues/984), after a long search I've discovered that [line 260](https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L260) in the H5BP print media query breaks printing of Google Maps if your browser is too wide (or, more specifically, if the map div is wider than ~800px). I've seen this on several WebKit browsers; example below tested on Chrome 46.

The map will look fine on screen, but when you hit print you'll notice that all of the tiles are missing. I've created [a demonstration here](https://github.com/rdlugosz/h5bp-google-maps-issue/blob/master/index.html) . Note that the only file changed in this repo from the default H5BP install is index.html.

The simple fix is to override the style via:
```
@media print {
  img {
    max-width: none !important;
  }
}
```

... but this took me a rather long time to discover (on and off for weeks until finally digging in with process of elimination). Is there a reason why this style is being set?

It strikes me that this could be considered a Google Maps bug and not that of H5BP, but thought this would be a better place to start. After all, a stock browser does not exhibit this problem until H5BP is introduced.

rdlugosz:
Following the lead of Twitter Bootstrap (see: https://github.com/twbs/bootstrap/pull/19000), remove the print media query that sets a max-width of 100% to images. This code was meant to prevent images from being cut off in print, but has a side effect of causing map tiles to disappear when printing medium-to-large maps (e.g., with Google Maps).

For reference, this style was added to H5BP in https://github.com/h5bp/html5-boilerplate/commit/b2faa1fc3cff20568848c98e9a51ecda61f3bf73

Fixes #1741

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -256,10 +256,6 @@ textarea {
         page-break-inside: avoid;
     }
 
-    img {
-        max-width: 100% !important;
-    }
-
     p,
     h2,
     h3 {


roblarsen:
Thanks!

Issue #1524, opened 2014-03-26
Drop IE8 support


alrra:
Microsoft is [ending the extended support](http://www.microsoft.com/en-us/windows/enterprise/end-of-support.aspx) for [Windows XP](http://en.wikipedia.org/wiki/Windows_XP#Support_lifecycle) on April 8, 2014. Since IE8 is mostly a Windows XP browser, and it's market share is slowly but steadily going down¹ , I think we should __start the discussion__ on dropping IE8 support.

--

¹  Some current estimates (feel free to share your own/other):
* [6.15% - StatCounter (March 2014)](http://gs.statcounter.com/#browser_version-ww-monthly-201403-201403-bar)
* [4.24% - W3Counter (February 2014)](http://www.w3counter.com/globalstats.php)
* [5.29% - Clicky (most recent)](http://clicky.com/marketshare/global/web-browsers/internet-explorer/)
* [6.93% - 8.73% - Wikipedia (as of September 2013)](http://en.wikipedia.org/wiki/Internet_Explorer_8)

--

Changes involved with dropping IE8 support:

* Remove IE8 related code
* Switch to jQuery `2.x.x`
* Update conditional comments for _"browse happy"_ prompt
* Remove html5shiv?
* Update documentation
* Other (which?)

robinio1:
I got a problem with this site http://www.fensternachmass-shop.de and IE, it won´t show the pictures.

davidjbradshaw:
Update as of June 1st.

* W3Counter: No longer listed in top 10 
* Clicky: 1.99%
* Statcounter: 1.95%
* Wikipedia: 1.55% (March)

arthurvr:
Wow. It's going in the good direction :)

Marsinator:
I don't know how I became a part of this conversation, you guys probably
have the wrong mail address.

Frank

Am Sonntag, 7. Juni 2015 schrieb Arthur Verschaeve :

alex:
(Not sure how I got CC'd on this issue, but since I'm here anywhere :-)). If you feel like looking at another source for browser usage, there's https://analytics.usa.gov/: sadly IE8 is still at 2.6%.

patrickkettner:
@Marsinator you commented on it back in December - https://github.com/h5bp/html5-boilerplate/issues/1524#issuecomment-67084713

@alex you were @ mentioned by julienetie back in December as well 

[...]

PR #1884, opened 2016-08-09
Resolves closed Greenkeeper PRs

caraya:
Resolves #1870, #1869, #1868 and #1864

jQuery is upgrading from version 1 to version 3. It may be a good idea to point to the upgrade guide somewhere to people know where to look in case something breaks.


coliff:
Changing H5BP to use/recommend jQuery 3 means that IE8 is no longer supported though. If that is the case it should very clearly state that - or perhaps there should be a major new version of H5BP (v6) which drops all support for IE8. (Support for IE7 was dropped with v5 of H5BP.)

caraya:
I'm ok reverting the commit and changing it back to a version 2. I'll wait to see what committers say on the matter

coliff:
Hi again. Yep I saw  #1524. So for now H5BP should use jQuery 1.12.4 which is the latest IE8 compatible version. jQuery 2.x doesn't support IE8 at all.

caraya:
moved back to jQuery 1.12.4 while waiting for further discussion on dropping IE8 support

roblarsen:
@caraya Yeah, the jQuery dev dependency is what's pushed into the index.html, so upgrading to 3.* shouldn't happen in this commit.  #1524 will handle that and everything else we need to do to finally unshackle ourselves from the deep past. As for the other upgrades in this PR does anything break?  If not, squash the two commits and create a commit message that describes the changes (what you're upgrading.). Since it will likely run long, put the issues that it closes in the body of the commit message. Then LGTM +1

caraya:
The last commit to this PR upgraded jQuery to 1.12.4. It's good to go

roblarsen:
Yeah, saw that. Thanks! You should still [http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html](squash) the commits and update the commit message to something that describes the changes you're making.

caraya:
No, tested the commits by running Gulp tasks that used the plugins and couldn't break it. 

Something weird happened when I rebased and it caused two commits.  Sorry, still learning how to rebase

roblarsen:
@caraya It's a little weird if you haven't done it. If you can't figure it out, feel free to create a fresh PR with just the required changes and close this one.

PR #1890, opened 2016-08-17
Re-add `sudo: false` to Travis config

scriptdaemon:
`sudo: false` is only default in repositories added in 2015 and later.

Ref: https://docs.travis-ci.com/user/ci-environment/


roblarsen:
@scriptdaemon Can you give me a two sentence description of what the difference is in this context?

scriptdaemon:
@roblarsen For repositories enabled on Travis before 2015 (e.g. HTML5 Boilerplate), `sudo: required` is still the default. `sudo: false` still needs to be used to enable the (theoretically faster) container-based environment.

I believe this was removed by mistake by someone who thought `sudo: false` was now the default for all repositories.

roblarsen:
Thanks!

PR #1891, opened 2016-08-17
Change lang attribute html element. Close #1879

### LINKED ISSUE #1879: Revisit lang attribute
I maintain the [W3C HTML Checker](https://validator.w3.org/nu/) and in looking at statistics logged  by it I notice there are a significant number of sites built from HTML5 Boilerplate that have `<html class="no-js" lang="">`. That is, they have the empty string as the value for the `lang` attribute, even when the document is clearly in English or another identifiable language.

I think this problem could maybe be prevented if HTML5 Boilerplate instead used something like `<html class="no-js" lang="CHANGE_ME">` rather than `<html class="no-js" lang="">`.

caraya:
added CHANGE as the value of the lang attribute for the HTML tag and a comment suggesting they users make the change. Closes  #1879

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,5 +1,6 @@
 <!doctype html>
-<html class="no-js" lang="">
+<!-- change value of lang attribute to the primary language of your content -->
+<html class="no-js" lang="CHANGE">
     <head>
         <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


roblarsen:
Hi @caraya thanks for your pull request. I'm going to close this now and see what comes out of the discussion in #1879

PR #1914, opened 2017-01-03
Set referrer policy to enable Referer headers on HTTPS

barraponto:
HTTPS served websites default to a Referrer Policy of `no-referrer-when-downgrade`, meaning no `Referer` header is sent at all when following a link to a non-HTTPS URL (see [specs][1]). This hurts analytics on the referred website and could hurt any claim of traffic generation from the referring site. It is then [recommended][2] to just use `origin-when-cross-origin` to send a stripped URL (no path, no queries) in `Referer` headers to cross-origin requests (full URLs are still used for same-origin). See more details in the [specs][3].

[1]: https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer-when-downgrade 
[2]: https://https.cio.gov/faq/#how-can-an-https-site-keep-sending-referrer-information-to-linked-http-sites?
[3]: https://www.w3.org/TR/referrer-policy/#referrer-policy-origin-when-cross-origin

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,6 +6,7 @@
         <title></title>
         <meta name="description" content="">
         <meta name="viewport" content="width=device-width, initial-scale=1">
+        <meta name="referrer" value="origin-when-cross-origin" />
 
         <link rel="apple-touch-icon" href="apple-touch-icon.png">
         <!-- Place favicon.ico in the root directory -->


roblarsen:
@barraponto Thanks for the PR and sorry for the slow (glacial!) response. I usually try to respond immediately because otherwise (as happened here) it can get lost in the shuffle. I'm hesitant to set this as a default because it might have a negative impact in some situations. I think it would be better documented as a section in [extend.md](https://github.com/h5bp/html5-boilerplate/blob/master/src/doc/extend.md). I'd love to hear some other opinions. I think it's important, so I want to make sure we do something with it. I just want to make sure we present it in the right way.

barraponto:
@roblarsen my point is that the default browser choice already has negative impact on its own.
Do you know any scenario where changing to `origin-when-cross-origin` would have (more) negative impact?

roblarsen:
@barraponto Sure, but it's the default. If we do nothing, that's what people are going to get. Once we make a choice it's going to change the default behavior for many, many authors. Since this attribute will go unchanged in a lot of cases, _any_ instance where the browser default or one of the more restrictive policies would be more appropriate would have more negative impact than the default.  This is _by definition_ a sensitive matter so opting people in to leak more information isn't the best option for us. 

I think the best compromise between promoting the `referrer` keyword (for people who want to be both less and _more_ restrictive in what they send) and associated values and not having unintended consequences would be to add the meta with [an empty string](https://www.w3.org/TR/referrer-policy/#referrer-policy-empty-string) assuming that browsers in fact respect the specification and fall-back to the default with an empty string set.

coliff:
I agree that this PR shouldn't be merged for the reasons @roblarsen has stated and it'd be best for the referer header to be set as a HTTP header rather than in the HTML page.

Shall we close this PR - and if necessary discuss adding documentation regarding referrer policy on one the H5BP server-configs repos (https://github.com/h5bp/server-configs) ?

roblarsen:
@coliff I think that's a great plan.

PR #1920, opened 2017-03-10
Deprecated clip: add clip-path as PE

ffoodd:
`clip` [is deprecated since 2014](https://www.w3.org/TR/css-masking-1/#clip-property). It works great but already has a more future proof version: `clip-path`.

@ryuran [suggested](https://twitter.com/ryuran78/status/778943389819604992) the `clip:path: inset(50%);` notation as it's the shorter possible to mimic our old clip.

Here's [a codepen](http://codepen.io/ffoodd/pen/gwKZyq?editors=1100#) and [a post translated by Hugo Giraudel](http://hugogiraudel.com/2016/10/13/css-hide-and-seek/).

Hope it'll help :)

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -128,6 +128,7 @@ textarea {
 .visuallyhidden {
     border: 0;
     clip: rect(0 0 0 0);
+    clip-path: inset(50%);
     height: 1px;
     margin: -1px;
     overflow: hidden;
@@ -146,6 +147,7 @@ textarea {
 .visuallyhidden.focusable:active,
 .visuallyhidden.focusable:focus {
     clip: auto;
+    clip-path: none;
     height: auto;
     margin: 0;
     overflow: visible;


roblarsen:
Thanks! I really appreciate the detailed doccumentation in the PR.

ffoodd:
Indeed, FWIW [Bootstrap has an opened PR about this](https://github.com/twbs/bootstrap/pull/25197) in which I suggested a workaround (see [this dedicated CodePen](https://codepen.io/ffoodd/pen/rpoqzd)).

Basically removing `clip-path` alongwith `clip` and use positionning to set hidden content offscreen.

I'm not able to test the perf scrolling issue, so if anyone dare to test this pen way for hiding content, it might be an easy fix.

tomasz1986:
@ffoodd Why change it at all though? Deprecated or not, `clip` does work perfectly fine in all web browsers starting from IE 4 and later. In terms of browser compatibility, there is really no comparison. In other words, why reinvent the wheel?

ffoodd:
It does not reinvent anything, since this PR is only meant to improve existing pattern, and my last comment suggests a very old and well known pattern.

However I do agree that `clip` is far more supported, and since we can't switch to `clip-path` without fallback we should start by reverting this PR.

FWIW performance issues in Chrome are just discovered, and aren't documented anywhere for now.

PR #1921, opened 2017-03-14
feat(performance): optimize pagespeed

jhabdas:
Optimize performance by leveraging the Fetch API and inlining critical path CSS. Also puts the Google Async script into the head of the document (because it's async).

Rundown of the changes:

- Use Fetch API for async dependency management (via [fetch-inject](https://www.npmjs.com/package/fetch-inject))
- Use feature detection to provide backwards compatibility
- Split main CSS file into critical and non-critical files
- Inline @necolas et. al CSS along with main in document head
- Remove SubResource integrity checking (nothing stops a hacker)

PageSpeed scores over LTE served with SimpleHTTPServer and Ngrok:

- Before (with Fetch support): Mobile (80) Desktop (89)
- After (with Fetch support): Mobile (88) Desktop (96)
- Before (without Fetch): UNKNOWN
- After (without Fetch): UNKNOWN


jhabdas:
Read the contributing guidelines prior to PR and did the best I could to keep this PR as succinct and focused as possible. The SRI stuff got in the way and wasn't being used consistently, so I pulled it using my good judgement.

roblarsen:
Thanks for putting together a clear PR.  At first blush I don't think this particular set of changes is right for HTML5 Boilerplate, but I'll leave the PR open if anyone else wants to chime in.

jhabdas:
@roblarsen thanks for your input. is there something specific you felt was out of place? message on the tin says:

> building fast, robust, and adaptable web apps or sites.

And this change helps with both _fast_ and _adaptable_ as it tees the boilerplate up for modern browser behavior leveraging the [Fetch API](http://caniuse.com/#feat=fetch)

roblarsen:
The simplest answer is that it's a downstream optimization. Yours is one of many ways this base code could be manipulated in order to produce a faster result. I want to preserve as many paths from here (this repo) to there (optimized production code that inherits from this repo) as I can.

jhabdas:
> I want to preserve as many paths from here (this repo) to there (optimized production code that inherits from this repo) as I can.

Thanks for the additional context, @roblarsen. I'd love to help H5BP meet its vision, so if there's anything in in the PR which maintainers feel helps achieve product vision I'd gladly make adjustments. Here are three patterns I feel this boilerplate should start following in order to achieve balance between speed, accessibility and maintainability:

- Async loading of nonessential scripts
- Async loading of non-critical CSS
- Movement of async scripts into the document head

I'll pause here while we wait for other maintainers to chime in.

jhabdas:
Thanks for looking.

PR #1934, opened 2017-04-07
Following the template in humanstxt.org

ajsb85:
http://humanstxt.org/Standard.html


QWp6t:
https://github.com/h5bp/html5-boilerplate/pull/1225#issuecomment-9005092

Issue #1605, opened 2014-09-23
Move Modernizr to the bottom


paulirish:
It has more potential to introduce unexpected situations, however it's much better for the user to just have no scripts up in the head at all.
## [Guide to Using Modernizr responsibly.](https://github.com/Modernizr/Modernizr/issues/878#issuecomment-41448059)

I bet >75% of sites dont need it in the head. I'd rather change this default and educate the 25% than watch as we slow down all these sites.

arthurvr:
@drublic Do you already have (some?) results?

drublic:
We don't have a lot of numbers from our production websites yet but here are some test numbers:

We now load html5shiv in `header.min.js` and our Modernizr custom build in `core.js`. Before the change, Modernizr was loaded in `header.min.js` too.

The change saves us in `header.min.js`
gzipped 3.4KB (2.7 KB compared with 6.1 KB before)
unzipped 10KB (5.6 KB compared with 15.6 KB before)

The saved data moves to `core.js`
gzipped 2.6 KB (82.4 KB compared with 79.8 KB)
unzipped 7 KB (258 KB compared with 251 KB)

This means we save bytes (because of how gzip works) and deliver the first view faster.
We did not encounter any problems with rendering within our project.

roblarsen:
ayo, I merged in code that does this 💯

PR #1940, opened 2017-04-21
Adds Modernizr 3

### LINKED ISSUE #1740: Modernizr 3.0
Modernisr 3.0 is out the door! Big round of applause! Awesome job!

H5BP is still running 2.8.3. This seems like a great time to take a look at the way Modernizr is used here. This is especially true since a lot of the activity on the Modernizr repo over the past few weeks is centering around getting people up and running with custom builds and (conversely) the idea of default builds.

So, hey! Discuss! What should the version of Modernizr that H5BP ships look like. Some standard set of detects? No detects at all? Does it ship Modernizr at all? Instead of a slimmed down Modernizr, why not just the HTML5shiv? What about a world with no IE8?

Related issues: #1605 (Move Modernizr to the bottom) and #1524 (Drop IE8 Support)

roblarsen:
Fixes #1740 and #1605

- removes the static custom build of modernizr
- moves Modernizr to the bottom
- builds a custom modernizr in gulp
- adds tests for the generated modernizr
- adds a modernizr config with new default tests


dreiquevada:
Hi. I wonder why modernizr has been moved to bottom. From what I know before, it speeds up performance when put in the `<head>` tags. Thanks! And sorry, I've not been updated much.

paulirish:
@dreiquevada 

deets on that:

https://github.com/Modernizr/Modernizr/issues/878#issuecomment-41448059
https://webperf.ninja/2015/be-mindful-with-modernizr/

dreiquevada:
I see. That makes sense to me now. Thank you Paul! Best regards, awesome work!

PR #1955, opened 2017-07-05
Improve ::selection documentation which became confusing after c7057be

### LINKED ISSUE #1953: Improve ::selection documentation
The comment `These selection rule sets have to be separate.` (https://github.com/maoberlehner/html5-boilerplate/blob/abc64efcb71847049620bf8ae78f3cd5ce26bf5e/src/css/main.css#L21) does not make any sense anymore in the source file after the changes made in c7057bef0c432c245d678d5ea4cf16526619cafb

I realize that it makes sense in the `dist/css/main.css` file tough. In my opinion the comment should not be in the source file and is also not necessary in the dist file. I'm happy to commit a pull request if the people are agreeing. Otherwise, how could this be solved so that the comment is added on build time?

maoberlehner:
After Autoprefixer was added a comment explaining that the vendor prefixed
`::-moz-selection` and the default `::selection` can't be combined became
obsolete. I (with the help of @roblarsen) rephrased the comment to be more
general and so that it still makes sense in the dist css file. A stackoverflow link
explains why vendor prefixed selectors can't be combined with other selectors.

Fixes #1953

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -18,7 +18,9 @@ html {
  * Remove text-shadow in selection highlight:
  * https://twitter.com/miketaylr/status/12228805301
  *
- * These selection rule sets have to be separate.
+ * Vendor-prefixed and regular ::selection selectors cannot be combined:
+ * https://stackoverflow.com/a/16982510/7133471
+ *
  * Customize the background color to match your design.
  */
 


roblarsen:
Thanks!

PR #1968, opened 2017-07-13
Editorconfig change 'will need a plugin' to 'may need'

coliff:
Visual Studio, BBEdit, WebStorm and others have built-in support for
Editorconfig files so we should change ‘will need to install a plugin’
to ‘may need to install a plugin’.


roblarsen:
@coliff You should open a documentation branch in your fork and push changes to that. We can squash all the documentation changes together before launch.

coliff:
OK sure. sounds like a good idea.

PR #1990, opened 2017-08-19
Re-adding position to .visuallyhidden class

svinkle:
Without `position: absolute` the `.visuallyhidden` element adds space to the container.

Demo: https://codepen.io/svinkle/pen/LjQJbZ

Thread: https://github.com/h5bp/html5-boilerplate/issues/1985#issuecomment-323485667


svinkle:
Never mind; this undoes the read order fix. Argh.

PR #1991, opened 2017-08-20
Restores stable visuallyhidden

roblarsen:
see the saga unfold #1985


nternetinspired:
Thanks @roblarsen 👍 

Current version generated additional whitespace, and your revision fixed that. Now working as expected :)

PR #1999, opened 2017-10-19
Add defer attribute to script tag in footer (index.html)

Utwo:
Add defer attribute to <script />

## Types of changes

## Checklist:

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -21,11 +21,11 @@
 
         <!-- Add your site or application content here -->
         <p>Hello world! This is HTML5 Boilerplate.</p>
-        <script src="js/vendor/modernizr-3.5.0.min.js"></script>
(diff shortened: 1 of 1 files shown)


coliff:
All the scripts are in the footer of the page - what's the need to add defer to them all?

Utwo:
Yes, because here are 2 or 3 scripts that need to load. We don't want to wait for the first one to download completely and only after that to load the second one. 
At least this is how I understand it.
More info here:
http://contentloaded.com/dom-ready/?head=style:500&body=html:1000;watch:1000;jquery:1000,defer;script:500,defer

mathiasbynens:
https://github.com/h5bp/lazyweb-requests/issues/42

tomasz1986:
What if the later scripts depend on the previous ones? If there are things that depend on Modernizr and jQuery, then they will break if the two libraries have not been loaded before.

Utwo:
Scripts are loaded async and run in order of appearance.

Utwo:
Yap, I agree!

PR #2022, opened 2018-02-04
Add redirection link & js to 404.html

arguiot:
Here is what has been added:
- a redirection link under the paragraph
- a script that will redirect to home page on clicking on the body

## Types of changes

## Checklist:

> It might but I couldn't find the `404.html` part.

### DIFF
--- a/dist/404.html
+++ b/dist/404.html
@@ -33,10 +33,16 @@
         }
 
         p {
-            margin: 0 auto;
+            margin: 5px auto;
             width: 280px;
         }
-
+		a {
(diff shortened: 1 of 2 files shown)


roblarsen:
Hi,

Thanks for your PR. I've reviewed it and I am not going to merge the code. 
1. Attaching a click event handler to the body would do unexpected things. I don't like the behavior to begin with and then as soon as you add any interactive element on 404 page, things would get strange. 
2. Linking to "/" won't work reliably.

PR #2030, opened 2018-03-19
Add .babelrc and .prettierrc to .gitattributes

frenzzy:
[Babel](https://babeljs.io/) and [Prettier](https://prettier.io/) are the most used front-end tools nowadays.


roblarsen:
Thanks!

PR #2036, opened 2018-04-04
HTTPS links

coliff:
Safer and faster

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -130,7 +130,7 @@ included in the project:
    ```
 
 4. Commit your changes in logical chunks. Please adhere to these [git commit
-   message guidelines](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)
+   message guidelines](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)
    or your code is unlikely be merged into the main project. Use Git's
    [interactive rebase](https://help.github.com/articles/about-git-rebase/)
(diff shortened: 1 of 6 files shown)


tomasz1986:
Safer, yes, but isn't HTTPS always slower than HTTP, especially on mediocre connections (e.g. mobile like 3G)? I am no expert in this field but from my experience and understanding there is quite a noticeable negative impact on how quickly websites load (particularly during the initial load, with no pre-cached files).

https://stackoverflow.com/questions/149274/http-vs-https-performance

I am not opposed to the change, only asking.

roblarsen:
If a site has implemented https, they're **hopefully** redirecting all http traffic to the secure version of that same resource. In that case, if we are pointing to an insecure resource and the server has to redirect the user that's always going to be _slower_ for the end user since they have to go through the redirect step. Skipping the redirection step by pointing directly at the secure resource is faster in that case, not matter what the difference is between http and https in terms of performance.

coliff:
Yep - HTTPS is important and advantageous. I like this site for explaining why: https://doesmysiteneedhttps.com/

In fact, maybe we should have a little section about HTTPS in the Extend section...

tomasz1986:
> If a site has implemented https, they're hopefully redirecting all http traffic to the secure version of that same resource.

Right, I was thinking about a situation where the same content is available through both HTTP and HTTPS. In this case everything is to go through HTTPS anyway. Thank you for a detailed explanation.

Amorn-Kingthong:
never mind

PR #2039, opened 2018-04-09
meta for responsive on mobile

pmaojo:
Make it responsive on mobile devices.


pmaojo:
I was in the wrong project. Sorry!

PR #2067, opened 2018-10-01
Added social cards support

arguiot:
Social cards support will increase the visibility of the website on social medias

## Types of changes

## Checklist:

Thanks!

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,12 +4,29 @@
 <head>
   <meta charset="utf-8">
   <meta http-equiv="x-ua-compatible" content="ie=edge">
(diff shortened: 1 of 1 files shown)


coliff:
hey there - thanks for the PR! I think adding Twitter Card and OpenGraph meta tags are important for many (most?) public-facing sites - but I think there's also many use cases where they are not needed. 

As such, I personally think that it's better to document this in our docs ([as we do now](https://github.com/h5bp/html5-boilerplate/blob/master/dist/doc/extend.md#social-networks)) rather than including it in the default HTML template.

Issue #1191, opened 2012-08-31
Drop Google's CDN for jquery.com?


PeterDeKeer:
In some corners of the web people push the microsoft cdn, here we push the google cdn.
Wouldn't it be better if we just used jquery.com for the jquery cdn?
And hope a lot of libraries follow suit?

PR #2122, opened 2019-03-18
jQuery CDN docs - remove unnecessary info

coliff:
I propose to remove this paragraph:

> The jQuery CDN version was chosen over other potential candidates
([like Google's Hosted Libraries](https://developers.google.com/speed/libraries/))
because it's fast ([comparable or faster than Google by some
measures](https://www.cdnperf.com/#jsdelivr,cdnjs,google,yandex,microsoft,jquery,bootstrapcdn/https/90))
and, (unlike Google's CDN) is available to China's hundreds of millions of internet users.
For many years we [chose](https://github.com/h5bp/html5-boilerplate/issues/1191)
the Google Hosted version over the jQuery CDN because it was available
over HTTPS (the jQuery CDN was not,) and it offered a better chance of
hitting the cache lottery owing to the popularity of the Google CDN.
The first issue is no longer valid and the second is far outweighed by
being able to serve jQuery to Chinese users.

We switched to jQuery CDN a long while ago and don't feel that providing this information is too useful. If you disagree and would rather keep it then instead we should just change this PR to change the last sentence to read '...serve jQuery to users in China' (rather than ...serve jQuery to Chinese users)

### DIFF
--- a/dist/doc/html.md
+++ b/dist/doc/html.md
@@ -155,6 +155,18 @@ the bottom of the page. A local fallback of jQuery is included for rare instance
 when the CDN version might not be available, and to facilitate offline
 development.
 
+The jQuery CDN version was chosen over other potential candidates
+([like Google's Hosted Libraries](https://developers.google.com/speed/libraries/))
+because it's fast ([comparable or faster than Google by some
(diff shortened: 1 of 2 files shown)


roblarsen:
@coliff I would prefer to leave the detail in here, just based on the amount of time it took to make the decision. It took forever. 

"users in china" is better language, however.

coliff:
cool - no prob. I've updated this PR as suggested.

PR #2128, opened 2019-05-24
update Apache Server Configs to 3.2.1

coliff:
REF: https://github.com/h5bp/server-configs-apache/blob/master/CHANGELOG.md

cc @LeoColomb

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "devDependencies": {
-    "apache-server-configs": "^3.1.0",
+    "apache-server-configs": "^3.2.1",
     "archiver": "^3.0.0",
     "babel-core": "^6.26.3",
     "babel-preset-env": "^1.7.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "eslint": "^5.16.0",
     "eslint-config-recommended": "^4.0.0",
     "eslint-plugin-mocha": "^5.3.0",
-    "glob": "^7.1.3",
+    "glob": "^7.1.4",
```
LeoColomb: Side note, might not be really important, but you're upgrading `glob` too.

coliff: yep, I should have mentioned this in the PR. We like to try and keep all the dev deps up-to-date. Updating this was a minor version and the tests passed and build was unchanged so it should be all good 👍 . Thanks for reviewing!

LeoColomb: > We like to try and keep all the dev deps up-to-date.

👍 

You can also use Dependabot for that, very handy (and just acquired by GitHub)! https://github.com/marketplace/dependabot-preview

coliff: YES! I actually have that installed on all my own GitHub repos - and found out about your update to Apache Server Configs from the fork I have of html5-boilerplate :-)  Yes, we should install Dependabot on all the H5BP repos I think. But they need to be tested before merging - and should ideally be hooked up to CI for testing.

roblarsen: We used to run greenkeeper and then that went crazy because the project was neglected for a few months in between active maintainers. I'm happy to look at something to automate the process.

PR #2153, opened 2019-09-04
remove .jscsrc config and remove gulp-jscs from package.json

coliff:
This is no longer needed since we migrated to eslint. (REF: https://eslint.org/docs/user-guide/migrating-from-jscs)

All tests pass and build completes.

## Types of changes

## Checklist:


coliff:
we could do, but there are lots of file extensions in there that we aren't using in this project - many of which are old unpopular formats such as RealAudio (.ra) and XHTML (.xhtml)... so I thought it might not make sense to remove just one.

PR #2218, opened 2020-04-14
Removing Browser Upgrade Prompt

### LINKED ISSUE #2210: Remove Browserupgrade prompt
**I'm submitting a ...**

- [ ] bug report
- [x] feature request
- [ ] other (Please do not submit support requests here (below))

HTML5Boilerplate default HTML includes a browser upgrade prompt which only displays on IE9 and below. I suggest that this should be removed in the next version of H5BP because:
- It doesn't display on IE10 (as IE10 doesn't support IE conditional comments)
- An IE8 user seeing the 'upgrade your browser' link does not provide supported alternatives. Chrome/Firefox/Opera are no longer supported on XP or Vista which an IE8 user would have.
- In my limited testing https://browsehappy.com/ doesn't actually load on IE8 anymore (HTTPS/SSL issue maybe?)
- [IE8/IE9 marketshare is tiny](https://caniuse.com/usage-table)

Instead, it'd possibly be a better idea for developers to detect version of IE in JavaScript (possibly by detecting the version of Windows too). If user is on Windows 7 or 8 - they should update to IE11 or download Edge. I think adding info or code for this is maybe out of scope of this project though, so suggest we simply remove the browserupgrade prompt.

roblarsen:
This commit removes the Browser Upgrade Prompt code we've had for oh so many years.

fixes #2210

### DIFF
--- a/dist/doc/html.md
+++ b/dist/doc/html.md
@@ -101,12 +101,6 @@ The central part of the boilerplate template is pretty much empty. This is
 intentional, in order to make the boilerplate suitable for both web page and
 web app development.
 
-### Browser Upgrade Prompt
-
-The main content area of the boilerplate includes a prompt to install an up to
-date browser for users of IE 9 and lower. If you intended to support IE, then you
-should edit or remove the snippet of code.
-
 ## Modernizr
 
 HTML5 Boilerplate uses a custom build of Modernizr.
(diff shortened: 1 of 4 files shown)


roblarsen:
That was a weird one. I hit it with a sledgehammer.

PR #2219, opened 2020-04-15
Sets anonymizeIp to true in GA snippet

### LINKED ISSUE #2212: AnonymizeIP for Google Analytics
@coliff writes:

>by default Google Analytics records the full IP address of a user visiting the site, but that full IP address is never available to the Google Analytics property admin. By anonymizing the IP address you can make your site more GDPR-compliant as a full IP address can be defined as publicly identifiable information. There is almost no downsides to this, other than the location of users can be slightly less accurate at a city-level. We already provide info about AnonymizeIP in the docs but I think it might be worthwhile including by default. Worth a discussion....

roblarsen:
closes #2212

### DIFF
--- a/dist/doc/html.md
+++ b/dist/doc/html.md
@@ -182,6 +182,12 @@ included.
 We use `analytics.js` rather than the newer `gtag.js` as
 [it's faster and supports tasks and plugins](https://github.com/philipwalton/analyticsjs-boilerplate/issues/19#issuecomment-333714370)
 
+Starting with version 8 we, by default, anonymizse IP addresses. By default
+Google Analytics records the full IP address of a user visiting the site, but
+that full IP address is never available to the Google Analytics property admin.
+By anonymizing the IP address you can make your site more GDPR-compliant as a
+full IP address can be defined as PII (publicly identifiable information.)
+
 The beacon transport mechanism is used to send all hits [which saves HTTP requests and improves performance](https://philipwalton.com/articles/the-google-analytics-setup-i-use-on-every-site-i-build/#loading-analytics.js).
 
 Google recommends that this script be placed at the top of the page.
--- a/dist/doc/html.md
+++ b/dist/doc/html.md
@@ -176,11 +176,12 @@ included.
 We use `analytics.js` rather than the newer `gtag.js` as
 [it's faster and supports tasks and plugins](https://github.com/philipwalton/analyticsjs-boilerplate/issues/19#issuecomment-333714370)
 
-Starting with version 8 we, by default, anonymizse IP addresses. By default
-Google Analytics records the full IP address of a user visiting the site, but
+Starting with version 8 we, by default, <a href="https://support.google.com/analytics/answer/2763052?hl=en">anonymizse IP addresses</a>.
+By default Google Analytics records the full IP address of a user visiting the site, but
 that full IP address is never available to the Google Analytics property admin.
 By anonymizing the IP address you can make your site more GDPR-compliant as a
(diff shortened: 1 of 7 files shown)


dist/doc/html.md
```diff
@@ -176,6 +176,13 @@ included.
 We use `analytics.js` rather than the newer `gtag.js` as
 [it's faster and supports tasks and plugins](https://github.com/philipwalton/analyticsjs-boilerplate/issues/19#issuecomment-333714370)
 
+Starting with version 8 we, by default, <a href="https://support.google.com/analytics/answer/2763052?hl=en">anonymizse IP addresses</a>.
```
coliff, line 179: anonymize typo here 🤓 
Also - should use markdown style links, not HTML ones I think

dist/doc/html.md
```diff
@@ -176,6 +176,13 @@ included.
...
+By default Google Analytics records the full IP address of a user visiting the site, but
+that full IP address is never available to the Google Analytics property admin.
+By anonymizing the IP address you can make your site more GDPR-compliant as a
+full IP address can be defined as PII (publicly identifiable information.)
```
coliff, line 183: sorry another thing... my fault as this was my comment... but this should actually be personally identifiable information (PII)
REF: https://en.wikipedia.org/wiki/Personal_data

src/doc/html.md
```diff
@@ -176,11 +176,11 @@ included.
 We use `analytics.js` rather than the newer `gtag.js` as
 [it's faster and supports tasks and plugins](https://github.com/philipwalton/analyticsjs-boilerplate/issues/19#issuecomment-333714370)
 
-Starting with version 8 we, by default, <a href="https://support.google.com/analytics/answer/2763052?hl=en">anonymizse IP addresses</a>.
+Starting with version 8 we, by default, [anonymize IP addresses](href="https://support.google.com/analytics/answer/2763052?hl=en).
 By default Google Analytics records the full IP address of a user visiting the site, but
 that full IP address is never available to the Google Analytics property admin.
 By anonymizing the IP address you can make your site more GDPR-compliant as a
-full IP address can be defined as PII (publicly identifiable information.)
+full IP address can be defined as PII (personally identifiable information.)
```
coliff, line 184: sorry, last nit-pick and then we're good. There's 2 line breaks after this section and should be just 1 to be consistent.

roblarsen: nit-picking is why we're here 👍

TheDancingCode:
Another level of privacy might be to respect the DoNotTrack signal of browsers.

In that case, the code would be wrapped in a conditional:
```js
var dnt = navigator.doNotTrack || window.doNotTrack || navigator.msDoNotTrack;
if (dnt !== 'yes' && dnt !== '1') {
    window.ga = function () { ga.q.push(arguments) }; ga.q = []; ga.l = +new Date;
    ga('create', 'UA-XXXXX-Y', 'auto'); ga('set', 'anonymizeIp', true); ga('set','transport','beacon'); ga('send', 'pageview');
}
```

coliff:
thanks @TheDancingCode - the DoNotTrack  has actually been deprecated now.

> DNT is not widely adopted by the industry, with companies citing the lack of legal mandates for its use, as well as unclear standards and guidelines for how websites are to interpret the header. Thus, it is not guaranteed that enabling DNT will actually have any effect at all. The W3C disbanded its DNT working group in January 2019, citing insufficient support and adoption. Apple discontinued support for DNT the following month.

https://en.wikipedia.org/wiki/Do_Not_Track

PR #2220, opened 2020-04-16
updating some markdown formatting

roblarsen:
TODO: line lengths

### DIFF
--- a/src/doc/usage.md
+++ b/src/doc/usage.md
@@ -27,7 +27,7 @@ HTML5 Boilerplate is a starting point, not a destination.
 
 A basic HTML5 Boilerplate site initially looks something like this:
 
-``` bash
+```
 .
 ├── css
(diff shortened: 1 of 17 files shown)


src/doc/usage.md
```diff
@@ -27,7 +27,7 @@ HTML5 Boilerplate is a starting point, not a destination.
 
 A basic HTML5 Boilerplate site initially looks something like this:
 
-```
+``` bash
```
coliff, line 30: is there any advantage/benefit to adding bash here? Does it change the color formatting in some way? If not, I think we could leave it out...

coliff:
Cool. I think this is an improvement. 👍

src/doc/usage.md
roblarsen: 👍

PR #2221, opened 2020-04-18
Fix for w3c title must not be empty rule

arnaud314315:
This pull request fixes w3c validator error - title must not be empty - by defining a default Hello title for the index.html page.


roblarsen:
Thanks for your submission. This isn't an appropriate change for this project. It's expected that end users of this project will supply a title.

Issue #2230, opened 2020-05-03
Quick start with npm init / npx


vltansky:
**I'm submitting a ...**

- [ ] bug report
- [x] feature request
- [ ] other (Please do not submit support requests here (below))

Inspired by [craete-react-app](https://github.com/facebook/create-react-app) I made small POC for html5-boilerplate. [Here is the video](https://skr.sh/v/030520/0vh6sRfA.mp4), and [repo](https://github.com/vltansky/create-html5-boilerplate).

![image](https://user-images.githubusercontent.com/5851280/80915533-85875280-8d5b-11ea-97f0-47dc6a3f39a1.png)

You can try it now with `npm init html5-boilerplate` or `npx create-html5-boilerplate`.
Repo without tests, validation and etc, so its just a POC don't take it seriously.

vltansky:
@roblarsen can you take a look on this?
I would like to prepare another version of it (different approach) and make a pull request if you will find this interesting.

roblarsen:
👍

roblarsen:
I've never authored for this infrastructure- can you outline what you would do differently? In this version is the repo name meaningful? Meaning, we have to have that pattern (create_-project-name_)

This looks like a fine starting place to me as it's just web-scale plumbing. We'd have to pull the repo into this org, of course and scaffold out some improvements to make it more of a proper project, but if this works, it's something we can start with right away (We can call it beta or alpha or whatever for the v8.0 release.)

roblarsen:
@vltansky can you follow up on this when you have time? I'm doing one other task for this release, but at this point it's just documentation and this feature left for the v8 release.

vltansky:
* npm package must have the prefix `create-` e.g `create-html5-boilerplate`, then npm recognize it as an ["init" script](https://docs.npmjs.com/cli/init) (`npm init html5-boilerplate`)
* Repo: Maybe better to manage it as a monorepo? Anyway, I've transferred the original repo to h5bp. 
* Npm package: I've added you as a maintainer, as I understood [this is the way to transfer ownership](https://docs.npmjs.com/transferring-a-package-from-a-user-account-to-another-user-account). (remove me from maintainers if needed)

vltansky:
Here is my suggestion on how this should work:

`npm init html5-boilerplate [folder] [--version | alias: -v]=<number>`
`folder` if no value provided will extract to the current folder (where console opened), otherwise, it will extract to the provided folder (create it if not exists).
[...]

PR #2229, opened 2020-05-02
feat(build): parcel (#2227)

vltansky:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

closes #2227

Issue #2227, opened 2020-05-01
Include bundler


vltansky:
**I'm submitting a ...**

- [ ] bug report
- [x] feature request
- [ ] other (Please do not submit support requests here (below)

Not sure if necessary, so it's more like a question/discussion, but in my point of view if I would want to make some static HTML page using html5-boilerplate I would definitely use bundler for my JS and CSS (probably would go for SCSS).
So I would use webpack or parcel to be able to use modern javascript and etc. So it would be comfortable for me if html5-boilerplate includes default webpack configuration, so I'll be able right away just to `npm install` and `npm run build`.

What do you think? Or maybe I miss the point of the html5-boilerplate and its not relevant?

roblarsen:
Many years ago, before there were JavaScript-based build tools of any stripe if you can imagine a time like that, we shipped with build tooling. We stripped that out to its own project in order to streamline the core offering. With the rise of Grunt/Gulp/WebPack that project was mothballed a few years ago. 

I'd be open to discussing something like this. I don't know what shape it would take but I'm open to it. I like the idea of something unobtrusive that could still be educational

vltansky:
What if html5-boilerplate would contain `package.json` with dev dependencies and scripts.
If developer needs bundler he can just `npm install` and then will be able to use `npm run build` / `npm start`.
Config file can be a part of html5-boilerplate or generated in post/pre npm install hook.
Although `parcel` claims to be zero config bundler, so it could be a good option..

vltansky:
Take a look on [example](https://github.com/vltansky/html5-boilerplate) with parcel.

roblarsen:
This is interesting. I think the idea of a stub package.json is potentially a good one. There's a lot there that we could stub out to illustrate and document.  @coliff what do you think? None of this is very obtrusive, but I think it opens up some possibilities. I've got a little bit to get done with GitHub Actions before this release can go live so there's time to work on this and expand on it a little.  

🤔

roblarsen:
Okay, I need to finish [shipping main.css v2.1.0](https://github.com/h5bp/main.css/issues/57) integrate that here and then I'm going to take a good look at this. The more I think about this the more I like it by itself (it's helpful, it should be unobtrusive and only adds one file to dist) and it opens up loads of possibilities (an entry in the package.json to add jQuery back in, for example.) 

I think it's going to push the release out  couple of weeks, but that's fine. I've still got the time.

nioposiquit:
[...]

PR #2240, opened 2020-05-08
Release V8.0.0

roblarsen:
This is a tracking branch for v8 integrations. A few of the changes we'll be making for this are big enough that I want a clean place to be able to stage them before merging them into master.

### DIFF
--- a/dist/.gitignore
+++ b/dist/.gitignore
@@ -1,3 +1,6 @@
 # Include your project-specific ignores in this file
 # Read about how to use .gitignore: https://help.github.com/articles/ignoring-files
 # Useful .gitignore templates: https://github.com/github/gitignore
+node_modules
+dist
+.cache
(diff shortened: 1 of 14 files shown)


coliff:
hey Rob. I've got some docs tidy-up work PRs in mind... nothing major, just fixing some redirects and stuff. Should I make PRs from and to this branch?

roblarsen:
@coliff the draft feature is sweet. I can remove the WIP app since drafts is universal and will make more sense to a general audience. 

Yes, you can do docs into this branch. I need to document the work we did with #2227 / #2229 and the work we're going to do with #2230 but that will come later so you can flow your work in now.

roblarsen:
I think this is actually pretty good right now. I think we'll just have some docs updates to do and the changelog, etc. but I think the big interim changes I was envisioning are probably done at this point. I had a mind to try to add another feature, but I'm going to hold out until the create feature is stabilized. Merging these into master will be a stable merge.

PR #2259, opened 2020-06-03
Fix `package.json` bug in README.md

snuggs:
Got your back @roblarsen


coliff:
:-) this is fixed in PR #2257

PR #2341, opened 2020-09-24
Renames main.js to app.js

### LINKED ISSUE #2340: Rename main.js -> app.js + update documentation
re #2339

roblarsen:
closes #2340


coliff:
looks good Rob- I think you need to run `gulp build` so that the empty 'app.js' is in the dist directory. (and main.js deleted).

roblarsen:
@coliff 👀

coliff:
when I run `gulp build` it updates Modernizr to 3.11.3. it's updated to that version in `package.json` so it should really be the same in dist/ too.

I think we should set dependabot to ignore Modernizr updates - that should be updated manually as it requires a new build of H5Bp to be made. Otherwise we get issues like this... I'll look into that.

roblarsen:
Yeah, the latest modernizr is included here. We should look into it so that it's manual and not depdendabot driven, for sure. But this one was never updated.

src/doc/usage.md
```diff
@@ -35,7 +35,7 @@ A basic HTML5 Boilerplate site initially looks something like this:
 ├── doc
 ├── img
 ├── js
-│   ├── main.js
+│   ├── app.js
 │   ├── plugins.js
 │   └── vendor
 │       └── modernizr.min.js
```
sthiepaan, line 41: Normally in operating systems, code editors and/or _IDE_, folders are listed before files. In that case `vendor/` should be placed before `app.js`. Can we move folders before files on tree structure so it is listed properly?

sthiepaan (changes requested):
Looks promising! Please have a look at the suggested change(s). Thank you! ✌️

sthiepaan:
@coliff _dependabot_ have [ignored_updates](https://dependabot.com/docs/config-file/#ignored_updates) option which should solve that problem. The downside of doing this is that we need to be aware of bumping ignored package(s) manually 🤷‍♂️

roblarsen:
@sthiepaan Since Modernizr is actually shipped with this project, I already watch it actively.

roblarsen:
I'll merge this tonight (GMT -04:00)

PR #2346, opened 2020-09-28
Remove plugins js

### LINKED ISSUE #2345: Do we still need plugins.js ?
**I'm submitting a ...**

- [ ] bug report
- [ ] feature request
- [x] other (Please do not submit support requests here (below))

There is currently a ['plugins.js' file](https://github.com/h5bp/html5-boilerplate/blob/master/src/js/plugins.js) which [is intended to be used for plugins such as jQuery plugins](https://github.com/h5bp/html5-boilerplate/blob/master/src/doc/js.md#pluginsjs).

In 'plugins.js' there is a 'small script to avoid console errors in browsers that lack a console'. Is that still a useful thing to include? Is this needed for any of our supported browsers? If it's not needed then maybe if it makes sense to either remove this or, if it is needed then perhaps concatenate this file with app.js at build time.

roblarsen:
closes #2345


roblarsen:
(this is based off of #2341, just for clarity of history, I'll merge that first)

sthiepaan (commented):
Since we remove _jQuery_ Library from the project, shouldn't we also remove all the references across the whole project? 🤔 I can see them in:
- `.github/`
- `dist/doc/`
- `src/doc/`
- `.htaccess`
- `gulpfile.babel.js`
- `humans.txt`

roblarsen:
@sthiepaan #2348

sthiepaan (approved):
LGTM

PR #2348, opened 2020-09-29
Jquery refs

roblarsen:
Follows in order behind the other two that are open. Not all references are important to remove, but some are.


src/doc/html.md
```diff
@@ -167,13 +167,11 @@ it before the other scripts in the bottom of the page:
 ```html
     <script src="js/vendor/modernizr-3.10.0.min.js"></script>
```
coliff, line 168: not related to this PR - but we should update modernizr version here.

coliff (approved):

sthiepaan (approved):
LGTM

Issue #2379, opened 2020-10-12
Comment-Out Google Analytics Boilerplate


TensorTom:
**I'm submitting a ...**

- [x] bug report
- [ ] feature request
- [ ] other (Please do not submit support requests here (below))

In the greater interest of privacy, many may consider it a bug of a boilerplate lib/repo to contain embedded active scripting from a remote Google property by default. Recommend simply commenting out the script tags with a message such as:

```html
<!-- Google Analytics: Uncomment the code block below & change UA-XXXXX-Y to be your site's ID. -->

<!--
<script>
    window.ga = function () { ga.q.push(arguments) }; ga.q = []; ga.l = +new Date;
    ga('create', 'UA-XXXXX-Y', 'auto'); ga('set', 'anonymizeIp', true); ga('set', 'transport', 'beacon'); ga('send', 'pageview')
</script>
<script src="https://www.google-analytics.com/analytics.js" async></script>
-->
```

hDmtP:
"**change UA-XXXXX-Y to be your site's ID.**" - how to be sure that one chooses a unique ID. 
And please can you explain how this scripting will help users privacy?

PR #2380, opened 2020-10-12
GA commented-out by default (#2379)

TensorTom:
## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.

Thanks!

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -29,12 +29,15 @@
   <script src="js/vendor/modernizr-3.11.3.min.js"></script>
   <script src="js/app.js"></script>
 
-  <!-- Google Analytics: change UA-XXXXX-Y to be your site's ID. -->
+  <!-- Google Analytics: Uncomment the code block below & change UA-XXXXX-Y to be your site's ID. -->
+
+  <!--
   <script>
-    window.ga = function () { ga.q.push(arguments) }; ga.q = []; ga.l = +new Date;
-    ga('create', 'UA-XXXXX-Y', 'auto'); ga('set', 'anonymizeIp', true); ga('set', 'transport', 'beacon'); ga('send', 'pageview')
+      window.ga = function () { ga.q.push(arguments) }; ga.q = []; ga.l = +new Date;
+      ga('create', 'UA-XXXXX-Y', 'auto'); ga('set', 'anonymizeIp', true); ga('set', 'transport', 'beacon'); ga('send', 'pageview')
   </script>
   <script src="https://www.google-analytics.com/analytics.js" async></script>
+  -->
 </body>
 
 </html>
--- a/src/doc/html.md
+++ b/src/doc/html.md
@@ -224,6 +224,12 @@ property admin. By anonymizing the IP address you can make your site more
 GDPR-compliant as a full IP address can be defined as PII (personally
 identifiable information.)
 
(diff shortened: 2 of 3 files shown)


coliff:
Thanks for the PR, though I don't think this should be merged. The full IP address is not sent to Google at all when using the `anonymizeIp` feature so the explanation in the docs update in this PR isn't quite right. It has a good explainer here: https://support.google.com/analytics/answer/2763052?hl=en. If we were to not enable Google Analytics in our html template by default, I think it'd be better to remove it altogether and document it in the docs rather than comment it out.

Also, PRs must include updated `/dist/` directory after making a build.

roblarsen:
Thanks for your contribution. Even PRs that don't get merged, like this, help us because it forces us to look at the proposed change and think about where we're at. I agree with @coliff I don't think this is a change we want to pull in.

TensorTom:
Looks like it was closed before I could reply. No matter. I care enough about privacy to reply anyway. I would have disagreed on the basis that the justification for not merging doesn't address the purpose of the merge and is also somewhat inaccurate. I read the docs fully before making the PR.

**Case 1:**

"The full IP address is not sent to Google at all when using the anonymizeIp feature."

This is not true and potentially misleading to new developers. From Google's docs:

"When a customer of Analytics requests IP address anonymization, Analytics anonymizes the address as soon as technically feasible at the earliest possible stage of the collection network."

This means that in fact, your full IP address was sent to Google. That you have already given your customer's/visitors IP addresses to them and are now taking it on faith of ethical good will that Google will always honour a policy. While some people are content to take Google at their word, call me crazy but I am not.

The first thing I did when I saw the GA code was remove it. You may disagree and believe Google's stated intent but I and many others don't. By leaving it on by default, you're saying that you believe most people (If asked) would prefer to have their customers' data sent to Google. I doubt this is the case.

**Case 2:**

Not only is it enabling a privacy concern, it's a needless one. I'd be willing to bet that most the pages that the boilerplate is used for never get connected to a Google Analytics account. This means that users are, by default, likely exposing their users to a privacy risk without even conferring any benefit. This leads directly into...

**Case 3:**

Even worse, this block of code is (by default) adding extra page load time, file-size, and an extra HTTPS connection upon page-load with no benefit to the developer or visitor (Most of the time). I can understand having it in a comment because that might be convenient for a subset of people. To leave it enabled by default though with so many downsides and (Most probably) hardly anyone actually using it makes little sense to me.

**Finally:**

I don't really care if you guys merge such a simple commit that took me all of a few minutes. I was surprised though that a commit with such an obvious technical benefit and zero downsides wasn't merged. Also, I did generate a new dist but I may have forgot to run it after updating the docs.

to each codebase their own I suppose

roblarsen:
Thanks for following up with that. I appreciate the thought you've put into this. To follow up I'll repeat what @coliff said and then add some color

> If we were to not enable Google Analytics in our html template by default, I think it'd be better to remove it altogether and document it in the docs rather than comment it out.

He said the above because, as a design decision that we made many years ago (maybe 8 or so), we don't ship anything extraneous. **We would never ship commented out code.** Proposing that we ship commented out code, no matter how well thought-out the reasoning is behind the change, is a non-starter. We would 💯 remove GA first.

PR #2384, opened 2020-10-18
Removed Google Analytics (GA)

jessaleks:

Reasoning:
a) Not everybody uses GA
b) It slows down page load which is a crucial metric (even if it's just 25ms on a 50kb/s network)
c) We have an obligation to respect users' privacy

While sure, a lot of people do use GA, not everybody does that. It's a bit unnecessary to always include every single time somebody uses this template. I have removed that from the project, though I do see a simple opt-in solution as enough.

As for the page load, ([https://wpostats.com/2018/06/13/zalando-revenue.html](https://wpostats.com/2018/06/13/zalando-revenue.html)) some tests show that seemingly unnoticeable changes have a noticeable impact on, financial metrics, even.

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -1,40 +1,30 @@
-<!doctype html>
+<!DOCTYPE html>
 <html class="no-js" lang="">
+  <head>
+    <meta charset="utf-8" />
+    <title></title>
+    <meta name="description" content="" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
 
-<head>
-  <meta charset="utf-8">
-  <title></title>
-  <meta name="description" content="">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta property="og:title" content="" />
+    <meta property="og:type" content="" />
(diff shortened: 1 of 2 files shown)


roblarsen:
Hi,

We're not going to remove this here. We will add it as an option to the create script. 

Thanks for your contribution.

jessaleks:
Ok. 

What is the reasoning, if I may ask?

roblarsen:
This PR (and many conversations like it) start from the supposition that Google Analytics is bad and that the project would be better without it. I don't think it's bad (I happily use it on my sites and have- just recently even- consulted on GA implementations over the years) and recognize that it's in use on hundreds of millions of sites. If you're not using some sort of analytics on your site, you're missing out on a big part of what it takes to make improvements to it over time.  Providing a good foundation for something that millions of people use and that I myself use is pretty straightforward for me.  I was happy to leave the PR open to see if it got some new discussion going, but it would take a lot for me to suddenly be in the "analytics is bad" camp.

PR #2412, opened 2020-11-26
Add missing period

mathiasbynens:


coliff:
Good spot. Thanks! We need to update the file in the src/ directory too though

coliff (approved):

PR #2459, opened 2021-02-13
Add a meta tag to opt out of cohort training

dmarti:
The upcoming "Federated Learning of Cohorts" proposal would enable data collection for identifying users who are members of particular groups based on their web history.

Sites will need to evaluate whether their content, or user patterns of activity on-site, can safely be used for this purpose. If this is the case, then this opt out can be removed.

More information and discussion are in the FLoC repo at: https://github.com/WICG/floc

## Types of changes

## Checklist:

Thanks!

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,10 @@
   <meta property="og:type" content="">
   <meta property="og:url" content="">
   <meta property="og:image" content="">
+  
+  <!-- Opt out of cohort data collection. Can be removed if site does not   -->
+  <!-- reveal sensitive info and privacy policy allows for cohort training. -->
+  <meta http-equiv="Permissions-Policy" content="interest-cohort=()">
 
   <link rel="manifest" href="site.webmanifest">
   <link rel="apple-touch-icon" href="icon.png">


coliff:
Sounds like an interesting idea - though as it’s in the proposal stage at the moment and isn’t implemented by any browsers/adtech companies I think it’s too early to consider adding it to the html5 boilerplate.

dmarti:
Thank you. There is a trial version of FLoC coming out in March 2021. (https://github.com/WICG/floc/issues/25)

I will post updates here as this proposal moves forward. If they decide to make cohort training opt-in for the site, then this meta tag will not be needed.

coliff:
Let's close this one. It's an interesting thing to keep track of, but it looks like it'll be something for the server configs rather than a meta tag.

roblarsen:
@coliff Sounds good. Are there associated issues over in the config repos? cc: @LeoColomb

LeoColomb:
FLoC is still planed to come after all? 😱
Yes, there was an issue for this https://github.com/h5bp/server-configs-apache/issues/266, but it's still not included per default.
Do you think it should be per default?

dmarti:
There is still an open issue in the FLoC project (https://github.com/WICG/floc/issues/13) covering HTML opt-out. The problem is that a `meta http-equiv=` tag can only set certain HTTP headers, and `Permissions-Policy` is not one of them. The FLoC project still needs to find a way to make an opt-out tag that would work for people who want to opt out but can't set the opt-out HTTP header.

dmarti:
@LeoColomb It looks like the existing approach to web server config files is to turn all the `Permissions-Policy` settings off, and then people can turn them back on as needed. (just looking at https://github.com/h5bp/server-configs-nginx/commit/36310b927b167ef8561dc6407ab9707038b804c7 ) If someone starts a web project now, there is a fairly good chance that it will be live at some point when FLoC-enabled browsers are out there. It probably depends on how you want to balance safety across different possible future scenarios against the bytes saved. (I expect that many people using h5bp as a starting point for their web projects will not be building an ad-supported site, so the advertising reasoning around FLoC might not apply to those projects.)

Issue #2487, opened 2021-03-25
Replace Parcel (or write a Parcel plugin to copy a list of static files)


roblarsen:
https://v2.parceljs.org/getting-started/migration

There are security alerts related to parcel, which we can't remediate with the existing (deprecated) v1 package. We should update it if the new version also allows us to solve #2474 If not, something else?

We have some open issues with main.css and create-html5-boilerplate as well, so maybe this is a proper release point coming together.

coliff:
I've been thinking about this too. Parcel v2's development appeared to have stalled a bit without an update since June 2020 - which made me hesitant to consider suggesting that we switch to it, but it [looks like there was a Beta 2 last week](https://github.com/parcel-bundler/parcel/releases). It's still 'Beta' though, so I'm not sure if that's something we should consider including with H5BP until it has a stable release.

There's https://github.com/vitejs/vite which could be considered too. I've personally not used it, but the feature-set looks great.

In any case, I'm keen to move away from using Parcel v1.

vltansky:
+1 for vite

Also vite have something similar to `create-html5-boilerplate` https://github.com/vitejs/vite/tree/main/packages/create-app

vltansky:
+1 for vite

Also vite have something similar to `create-html5-boilerplate` https://github.com/vitejs/vite/tree/main/packages/create-app

coliff:
Parcel v2 Release Candidate 1 has just dropped!
https://v2.parceljs.org/blog/rc0/
As I mentioned earlier in the year, I was hesitant to switch to Parcel v2 while it was still in Beta as there could be more breaking changes but now its at a Release Candidate stage it should be pretty safe to think about upgrading to it. It fixes a lot of issues and has many improvements.

roblarsen:
Oh, we should try it out, that would be great.

roblarsen:
#2581 drove me crazy, so at this point I'm voting we come up with a replacement for parcel. The stuff parcel does is great and would serve as a nice foundation for someone looking to dip their toes into modern app development. The thing is, I can't tell the stupid thing to copy files not referenced in the HTML without writing a plugin, so a bunch of files we would want to ship are not getting picked up in the build process. Deal-breaker, as far as I'm concerned. Hashing the favicon was pretty bad, but not even copying it is worse. 

The two options I see are:

1. write the stupid plugin to take a list of files and copy them
2. Use something more configurable and use it as an opportunity to create a baseline configuration for people to work from

PR #2502, opened 2021-04-11
parcel-bundler deprecated. Change to parcel v2

trasherdk:
## Types of changes

- [x] Update dependencies to fix security warnings

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [x] I have not changed any code

Thanks!


roblarsen:
Hi, we're already thinking about what to do with [parcel](https://github.com/h5bp/html5-boilerplate/issues/2487) so whatever solution we come up with will have some discussion attached to it. That aside, this PR is problematic, even if we decided to go with simply updating parcel. Deleting the `dist` directory and adding it to `.gitignore` is a non-starter as the `dist` directory is what we ship when we release a new version. 

Thanks for your contribution, however.

PR #2551, opened 2021-06-26
remove 'Common Prefetch Links' from extend docs

### LINKED ISSUE #2549: Suggest remove 'Common Prefetch Links' from docs
**I'm submitting a ...**

- [x] bug report
- [ ] feature request
- [ ] other (Please do not submit support requests here (below))

https://github.com/h5bp/html5-boilerplate/blob/master/src/doc/extend.md#common-prefetch-links

I don't think these are very helpful.

- Google's hosted libraries CDN isn't as popular as it once was.
- Users probably shouldn't be serving content directly from s3.amazonaws.com (it's not a CDN) and instead have CloudFront in front of it.
- Microsoft are actively discouraging users from using the Microsoft one. (https://docs.microsoft.com/en-us/aspnet/ajax/cdn/overview) and I agree with their statement: 'Production applications should not take a hard dependency on 3rd party CDN assets.'

It's easy enough for a user to add a preconnect themselves and there are already a couple of examples directly above: https://github.com/h5bp/html5-boilerplate/blob/master/src/doc/extend.md#explicit-prefetches

coliff:
Fixes #2549

Also removes old link about IE9 dns-prefetching (page last updated 10 years ago) and updates Smart App Banner URL to prevent redirect.

## Types of changes

### DIFF
--- a/dist/doc/extend.md
+++ b/dist/doc/extend.md
@@ -25,7 +25,7 @@ everything fits with everyone's needs.
 
 Stop bothering everyone with gross modals advertising your entry in the App
 Store. Including the following [meta
-tag](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/PromotingAppswithAppBanners/PromotingAppswithAppBanners.html)
+tag](https://developer.apple.com/documentation/webkit/promoting_apps_with_smart_app_banners)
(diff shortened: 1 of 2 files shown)


roblarsen (approved):
🚢

PR #2581, opened 2021-08-16
Upgrades Parcel

### LINKED ISSUE #2487: Replace Parcel (or write a Parcel plugin to copy a list of static files)

roblarsen:
Closes #2487 (although we should discuss whether or not we want to wait for the full release of 2.0.0)


coliff:
- I think .cache needs to be added to the `.gitignore` (no need for `.parcel-cache` anymore I think).
- It adds a dist directory within the `/dist` directory. seems a bit overly complicated having a src and a dist and having to build again from dist
- I don't like the hashed filenames and think we should disable that using a parcel config- especially for files such as normalize where the version number is already part of the filename. e.g. `modernizr-3.11.7.min.a04626e5.js` should just be `modernizr-3.11.7.min.js`.

roblarsen:
Copying static files not referenced in the html file requires we write a plugin, apparently. I'm voting we rip parcel out and do something else. I'll return to #2487 to discuss

PR #2636, opened 2021-11-11
Push releases to Template Repo

roblarsen:
Just picking up the work @vltansky was doing since I have some time right now. I've created the template repo, so it's just a matter of making sure this works and it a tidy user experience. The fundamentals are there, I think.


roblarsen:
This probably works 🤷 

I have it set to commits to main, for now. Once I see it working, I will switch the trigger to releases.

roblarsen:
I think I'm just going to merge this, because, while it's weird to merge code I haven't tested, there's really no easy way to do this otherwise 😿 I need to iterate on this, anyway.

PR #2637, opened 2021-11-12
Proposal: ship a simple custom element

roblarsen:
It also revises the modernizr build to reference only what we're using. There would need to be related documentation about both what we're doing with Modernizr here and, of course, the basics of a custom element.

## Types of changes

## Checklist:

- [ ] I have read the **[CONTRIBUTING](https://github.com/h5bp/html5-boilerplate/blob/main/.github/CONTRIBUTING.md)** document.

Thanks!


coliff:
I don't understand the purpose of this right now. What advantage does this have?
Custom elements are neat but I don't think every site needs them and don't know why an example should be included.

roblarsen:
> I don't understand the purpose of this right now. What advantage does this have? Custom elements are neat but I don't think every site needs them and don't know why an example should be included.

I'm just looking at different things right now. This project wasn't always as spare as it is now when it came to shipping example code, so this idea is an extension of the original approach (which had much more "boilerplate")

mr-vaibh (approved):

PR #2638, opened 2021-11-12
Proposal: Replace modernizr with modernizr-esm 

roblarsen:
This builds off of #2637  

This is a bit out there since modernizr-esm has like 20 stars, but... the idea of doing a custom element example and then testing against a `window.customelement` global is kind of a bummer. 

This is the way it should look in 2021 and the main Modernizr project is on life-support. Yet, the need still exists.


roblarsen:
👋 @niksy

Issue #2474, opened 2021-03-01
Parcel renames favicon.ico


joshbambrick:
**I'm submitting a ...**

- [ ] bug report
- [x] feature request
- [ ] other (Please do not submit support requests here (below))

## Notes:
I struggled for a while to understand why my favicon was not showing properly. It turns out I needed to add `<link rel="icon" href="favicon.ico" type="image/x-icon">` to `index.html` but this did not seem obvious to me and the comment `<!-- Place favicon.ico in the root directory -->` was not very helpful (in fact, I interpreted it as that I only need to update `favicon.ico` and not add the above line).

roblarsen:
We do have it documented and you interpreted it correctly. Browsers pick up favicon.ico in the root of your site. Something else was going on with your icon. 

https://github.com/h5bp/html5-boilerplate/blob/master/dist/doc/html.md#favicons-and-touch-icon

> Favicons and Touch Icon
The shortcut icons should be put in the root directory of your site. favicon.ico is automatically picked up by browsers if it's placed in the root. HTML5 Boilerplate comes with a default set of icons (include favicon and one Apple Touch Icon) that you can use as a baseline to create your own.

read this post by @mathiasbynens for full details

[rel="shortcut icon" considered harmful · Mathias Bynens](https://mathiasbynens.be/notes/rel-shortcut-icon)

Also see #2243

joshbambrick:
I would ask you to double check this.

I understand that browsers pick it up in the root of the website, however, as I understand it, parcel modifies the file name which means this no longer gets picked up. The file 'mysite/favico.ico' gave me a 404. Having added the above modification, desktop browsers picked up the new favicon, and I can see the file name as `favicon.b30494fb.ico` in my case.

roblarsen:
I don't need to double check the basic behavior or the way we implement favicons. The commit removing [<link rel="shortcut icon" href="/favicon.ico">](https://github.com/h5bp/html5-boilerplate/commit/6a71a7e9d51d3e936b50f7220426b6eca07b5512) is almost ten years old and Mathias' post is even older. So, what you're describing is not an issue with the documentation, but instead a configuration issue with Parcel, which we do provide as a convenience. We would not suggest adding `<link rel="icon" href="favicon.ico" type="image/x-icon">` back into the HTML. What we will do is ensure Parcel doesn't cache-bust the favicon.

joshbambrick:
Great, I think that would be very helpful.

[...]

PR #2641, opened 2021-11-17
Proposal: Let's Replace Parcel With Webpack

### LINKED ISSUE #2487: Replace Parcel (or write a Parcel plugin to copy a list of static files)

roblarsen:
Closes #2487 and #2474

This is just a development server at this point. It hot reloads HTML JS and CSS.  That's enough for me. I guess we'll need a build step. 

I also need to change the docs, if we go this way.


coliff:
Nice! I like idea of using Webpack instead of Parcel. Some quick feedback/things I noticed:

- the `src/package.json` isn't using latest versions - we should update them:
```txt
 html-webpack-plugin   ^3.2.0  →   ^5.5.0     
 webpack-cli           ^3.3.6  →   ^4.9.1     
 webpack-dev-server    ^3.7.2  →   ^4.5.0     
 webpack              ^4.36.1  →  ^5.64.1
 ```

- the `entry:` in the webpack config should be `app.js` not `main.js`

- the `webpack.config.js` indenting should be 2 not 4 to match `.editorconfig` rule and use single quotes not double quotes to match eslint config

roblarsen:
@coliff can you take a look at this? I want to chunk out the tasks a little bit and I'm interested in merging this part first, then look at builds for the dist folder and for the project itself.

coliff:
heya. so if a user clones this repo and does `npm install`, then changes directory to `src` and runs `npm run start` the server won't serve the CSS because those were copied to the dist dir instead. should a user run local server from dist instead?

Unrelated note, we should remove that `Place favicon.ico in the root directory` comment from the HTML as it is no longer needed (the meta tag is included)

coliff:
There's a couple of eslint 'errors' on `webpack.config.js` (when it's checked automatically with ESlint plugin in VS Code) but I assume they are fine.
image
Should be eslint ignore `webpack.config.js` - or add rule to turn off `no-undef` errors?

roblarsen:
@coliff Yeah, this end user webpack stuff needs to be run from `dist`.  I'm going to write up a bunch of documentation about how the `dist` folder *is* the project for end users to make that point clear (the addition of the template repo will help reinforce that.  That's one of the things that came out of that Chris Coyier article. 

I'll remove the comment here.

roblarsen:
> There's a couple of eslint 'errors' on `webpack.config.js` (when it's checked automatically with ESlint plugin in VS Code) but I assume they are fine. image Should be eslint ignore `webpack.config.js` - or add rule to turn off `no-undef` errors?

The errors aren't really errors, but we can turn off no-undef to make the red squiggles go away.

coliff:
Great. I'll open a PR for eslint config update, we also need to update the `"` to `'` to match the semi-colon rule in eslint itself!

roblarsen:
@coliff  I got the eslint stuff. I added node as an environment and changed the quotes.

roblarsen:
I need to update the docs since right now all this does is run the development server. We reference three npm scripts- there's just one now. I will create a build task later, but I don't want to have this PR open forever.

vltansky:
Why not Vite?

roblarsen:
> Why not Vite?

Webpack is something I use, so conceptually it was the easiest thing for me to get up and running. It also allows us to do the things that I require for this tool. It's also a giant, stable project at the heart of the web, so it seems safest.

coliff (approved):

PR #2655, opened 2021-12-02
Moving docs out of src and dist

roblarsen:


roblarsen:
Phase 1 of a documentation update. Next, more writing.

PR #2715, opened 2022-02-28
"About this repo" documentation

### LINKED ISSUE #2652: Update docs to focus on the difference between this repo and the project as end-users use it
We need to do a much better job of describing the difference between this project and the **output** of this project, which is the thing that people actually use.  

For example, the gulpfile is used by basically two people most of the time, me and @coliff (as well as any kind contributors who notice that we need to run a build 🥇) It's a tool used to **build** HTML5 Boilerplate, its not part of the thing that we ship. So when Chris Coyier pointed it out in an article last year it was like a bolt of lighting to my brain that we needed to reposition the way we present the repo. 

The thing we ship is the [dist](https://github.com/h5bp/html5-boilerplate/tree/main/dist) folder of this repo, the [npm package](https://www.npmjs.com/package/html5-boilerplate), the output of [create-html5-boilerplate](https://github.com/h5bp/create-html5-boilerplate) and, breaking news, the [template repository](https://github.com/h5bp/html5-boilerplate-template) we just launched.

I need to write those words in the readme. 

(not those exact words, but words like them. Just better)

roblarsen:
closes #2652

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,18 +36,11 @@ So, if you're looking for a quick start template to build a web site or
 application, look at the options in the
 [Quick Start section of this document](https://github.com/h5bp/html5-boilerplate#quick-start)
 
-If you want to help us _improve_ HTML5 Boilerplate then you can start with the documentation [here](.github/CONTRIBUTING.md), which includes steps to clone this repo and getting it set up for development.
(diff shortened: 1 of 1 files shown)


roblarsen:
@coliff 👁️

coliff (approved):

PR #2756, opened 2022-04-18
Create dependency-review.yml

coliff:
This adds a new GitHub Action.
https://github.com/actions/dependency-review-action

This action scans your pull requests for dependency changes and will raise an error if any new dependencies have existing vulnerabilities. Dependabot already warns about EXISTING vulnerabilities in the code base, with this PR merged we'll also be warned BEFORE they are merged.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,14 @@
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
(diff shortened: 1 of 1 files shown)


roblarsen (approved):
🦖

coliff:
in the **Settings > Branches** I think we should add Branch Protection Rules for the main branch and require the GitHub Actions to pass tests. I do this on all my repos and I think it's helpful.

e.g.
image

roblarsen:
@coliff Sure. We can lock it down better. I use strict setting for this stuff in other repos and have wanted to bring some of those best practices over here. Let me do some housekeeping.

coliff:
cool - and adding https://github.com/github/codeql-action#codeql-action too would be good.

PR #2815, opened 2022-07-19
Update .cspell.json with more false positives

roblarsen:
I'll try to catch them all with this PR

image


roblarsen:
This actually looks good.

coliff (approved):

PR #2831, opened 2022-08-03
Gulp esm

roblarsen:
closing #2824 which forces us to do something useful. Since ES modules are supported natively, we can get rid of babel and still use modern JS. 

This PR:

- removes babel as a dependency
- changes the gulpfile and the two test scripts to be `.mjs` All three were already using import, etc. so the upgrade was actually pretty easy. The specific change was to fix all three scripts to be able to read json files
- updates the eslintrc file to recognize the ability to do the above


roblarsen:
@coliff this is good to go, I think.

PR #2946, opened 2023-04-01
Removing initial-scale in HTML files and updating documentation

cevdetardaharan:
This PR removes the initial-scale attribute in HTML files and updates the documentation accordingly. The initial-scale attribute is unnecessary as its default value is already set to 1. This change fixes an issue and does not introduce any new functionality or breaking changes. I have followed the code style of the project, updated the documentation, and read the CONTRIBUTING document. Please note that pull requests should be thought of as a conversation, and I am open to any feedback or suggestions to ensure a clean merge.

## Types of changes

## Checklist:

- [x] I have read the **[CONTRIBUTING](https://github.com/h5bp/html5-boilerplate/blob/main/.github/CONTRIBUTING.md)** document.

Thanks!

### DIFF
--- a/dist/404.html
+++ b/dist/404.html
@@ -4,7 +4,7 @@
 <head>
   <meta charset="utf-8">
   <title>Page Not Found</title>
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="viewport" content="width=device-width">
   <style>
     * {
       line-height: 1.2;
--- a/dist/index.html
+++ b/dist/index.html
@@ -5,7 +5,7 @@
   <meta charset="utf-8">
   <title></title>
   <meta name="description" content="">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="viewport" content="width=device-width">
 
   <meta property="og:title" content="">
   <meta property="og:type" content="">
--- a/docs/html.md
+++ b/docs/html.md
@@ -59,7 +59,7 @@ the MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_met
 HTML5 Boilerplate comes with a simple setup that strikes a good balance for general use cases.
 
 ```html
-<meta name="viewport" content="width=device-width, initial-scale=1">
+<meta name="viewport" content="width=device-width">
 ```
 
 If you want to take advantage of edge-to-edge displays of iPhone X/XS/XR you
--- a/src/404.html
+++ b/src/404.html
@@ -4,7 +4,7 @@
 <head>
   <meta charset="utf-8">
   <title>Page Not Found</title>
-  <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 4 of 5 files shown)


roblarsen:
This kind of change would require a ton of testing. I no longer have access to browserstack so I can't actually run the tests required to pull in a change like this. It's one thing to say that the default is 1, but it's another to actually see what browsers do when you remove the explicit definition of this property.

coliff:
> This kind of change would require a ton of testing. I no longer have access to browserstack so I can't actually run the tests required to pull in a change like this. It's one thing to say that the default is 1, but it's another to actually see what browsers do when you remove the explicit definition of this property.

I agree with this. 

@cevdetardaharan have you seen this as a recommended best practise anywhere? All of the big projects I’ve seen still include the initial-scale part on the meta tag. As it is, I’m a bit reluctant to remove it.

cevdetardaharan:
Firstly, removing the attribute can help reduce unnecessary code and potentially improve performance. While the code savings may be small, every byte counts in today's mobile-first, speed-focused web development landscape.

Secondly, the default value of initial-scale is already 1, so including the attribute may not actually provide any additional benefit. While some older or alternative browsers may not use the default value, these browsers are becoming increasingly rare and should not be the primary focus of our development efforts.

Finally, while it's true that some frameworks and libraries may rely on the initial-scale=1 attribute being present, these issues can often be addressed through simple workarounds or documentation updates. Sacrificing code cleanliness and optimization for the sake of framework compatibility is not always the best decision.

In summary, removing the initial-scale=1 attribute can help reduce unnecessary code and potentially improve performance without sacrificing compatibility or causing significant compatibility issues. Thank you for your consideration and I hope you will support this pull request.

PS: It's worth noting that Internet Explorer is now considered a legacy browser and is no longer officially supported by Microsoft. According to recent data from [StatCounter](https://gs.statcounter.com/browser-market-share/desktop/worldwide), as of February 2023, only 0.64% of internet users globally are still using Internet Explorer. This means that the vast majority of users are on modern browsers that support the default value of initial-scale without issue. While we should always strive for compatibility and accessibility, it's important to prioritize optimization and modern web development practices in order to provide the best user experience possible.

roblarsen:
Those are very well thought out arguments. I'm still not pulling in a change like this without actually knowing what happens in browsers (including non-target browsers.) So if we can test this in basically everything and the results are acceptable, then this is an interesting change to make. I just don't pull the trigger on fundamental changes without knowing exactly what I'm doing.

cevdetardaharan:
Using a tool like Playwright to create a comprehensive testing environment that covers a wide range of browsers can help to ensure that the proposed change is thoroughly tested and that any potential issues are identified and addressed. It's important to note, however, that there may be limitations to what can be tested in this way, particularly with regard to non-target browsers. In such cases, it may be necessary to rely on other testing methods or to accept some degree of risk in making the change. So I think we should do a testing environment for that purpose!

Issue #2939, opened 2023-03-17
Drop Normallize.css


dgibson666:
### Prerequisites

- [X] I have [searched](https://github.com/h5bp/html5-boilerplate/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/h5bp/html5-boilerplate/blob/main/.github/CONTRIBUTING.md)

### Proposal

Since you're providing so much base CSS, much of which appears to overlap with a CSS reset, it's probably best to drop normalize.css as a dependency and include your own CSS reset. 

### Motivation and context

It's been a while since a new release here, but even longer for normalize.css. It was last updated in 2018 and still littered with rules for IE10 and 11. There's probably a lot that needs to be reevaluated in a post-IE world, and the CSS reset is certainly one of them.

I've been using Andy Bell's as a starting point (https://github.com/Andy-set-studio/modern-css-reset) but have not had the time to go through normalize.css rule by rule to verify what's still valid.

This is important work for modernizing projects.

I'd love to help where I can if this project is still alive.

roblarsen:
Happy to discuss! We're actually close to a new release. The webpack build PR I have open is the last major issue, but this is worth exploring.

roblarsen:
@coliff 👓

roblarsen:
So, I pulled in the Webpack build PR and will have some housekeeping to do over the next couple of weeks. To that end, if, over the next couple of weeks we want to discuss swapping out normalize that would be great. The work of swapping it out is pretty east.

wboswall:
Why doesn't the project just fork normalize and create its own version that way its a win-win? Just a suggestion

roblarsen:
That sort of work doesn't happen for free, basically. There's a significant time cost with every project we take on, so if there's not someone else managing it, it's not something I'm interested in doing.

wboswall:
@roblarsen Okay, that's fair.  What is the game-plan on normalize.css?

coliff:
I think we should remove Normalize altogether and don't replace it with anything... instead suggest to users that they consider using one of the many available CSS resets (perhaps with a link to: https://github.com/search?q=css%20reset&type=repositories)

dgibson666:
In my research I also ran across this one, which seems the best maintained:
https://github.com/u1ui/norm.css

[...]

PR #2960, opened 2023-05-05
Drop Normalize

roblarsen:
This is not actually decided, but since the work to do so would be extensive, I want to have it complete so that if we decide to move in that direction then we can do so easily. Please discuss! #2939


coliff:
Great stuff. Definitely makes sense to remove this. I think we could make a mention of this no longer being included and link to some potential replacements, but that can be done separately.

rgant:
Looks like this was missed: https://github.com/h5bp/html5-boilerplate/blob/87bdc80ab9e35dcb52d75ffc926d8ed67991c626/src/index.html#L19

ejgutierrez74:
Is there any link to new or more robust replacement ?
Im 9.0.1 no normalised.css, but no link or suggestion of replacement....

PR #3105, opened 2024-04-01
Updates for gulp 5

roblarsen:
## Types of changes

## Checklist:

- [ ] I have read the **[CONTRIBUTING](https://github.com/h5bp/html5-boilerplate/blob/main/.github/CONTRIBUTING.md)** document.

Thanks!

### DIFF
--- a/dist/css/style.css
+++ b/dist/css/style.css
@@ -1,4 +1,4 @@
-/*! HTML5 Boilerplate v9.0.0-RC1 | MIT License | https://html5boilerplate.com/ */
+/*! HTML5 Boilerplate v9.0.0 | MIT License | https://html5boilerplate.com/ */
 
 /* main.css 3.0.0 | MIT License | https://github.com/h5bp/main.css#readme */
 /*
--- a/dist/package.json
+++ b/dist/package.json
@@ -15,10 +15,10 @@
   },
   "devDependencies": {
     "copy-webpack-plugin": "^11.0.0",
-    "html-webpack-plugin": "^5.5.3",
-    "webpack": "^5.88.2",
+    "html-webpack-plugin": "^5.6.0",
+    "webpack": "^5.91.0",
     "webpack-cli": "^5.1.4",
-    "webpack-dev-server": "^4.15.1",
-    "webpack-merge": "^5.9.0"
+    "webpack-dev-server": "^5.0.4",
+    "webpack-merge": "^5.10.0"
   }
 }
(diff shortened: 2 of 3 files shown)


coliff:
Test failure:
image

roblarsen:
@coliff tests are passing here and in my codespace. Can you double check?

coliff:
Tested on Windows 11 before - get a different error on macOS
Untitled

[.DS_Store](https://en.m.wikipedia.org/wiki/.DS_Store) is in our .gitignore but should ideally set it to be excluded from dist dir too.

Koribraide:
> @coliff tests are passing here and in my codespace. Can you double check?

whats the issue ?

coliff:
> > @coliff tests are passing here and in my codespace. Can you double check?
> 
> whats the issue ?

npm run build passes, but npm run test does not on my Windows 11 PC. I tried a few things including deleting node_modules and reinstalling. Do you want to give it a try?

image

Koribraide:
hmmm, what is the result supposed to look like?

coliff:
> hmmm, what is the result supposed to look like?

Pass with no errors

roblarsen:
BREAKING NEWS

I just checked out the v9.0.0 tag and it's broken there. 

So... I don't know when this test last passed on Windows. I am actually on Windows as I type this but I've become a Codespaces devotee so I did the entire 9.0.0 release on GitHub and the PR tests are all running in similar environments. I'm going to backwards through tags to find out when it last worked at start from there. It definitely used to work because before Codepsaces everything I did went through Windows. 

The issue is that "css\style.css" is not "css/style.css" 

```
 'webpack.common.js',
  'webpack.config.dev.js',
  'webpack.config.prod.js',

  'css/', // for directories, a `/` character
  // should be included at the end
  'css/style.css',
```


```
 16) Test if all the expected files, and only them, are present in the build directories
       dist
         "css\style.css" should NOT be present:

      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:        

  assert(false)

      + expected - actual

      -false
      +true
````
 
@coliff I'd like to merge this PR and then I'll create a separate issue for the failing tests since they pre-date the Gulp upgrade.  Once I fix that we'll put out a patch release.

coliff (approved):

Koribraide:
@coliff fixed?

roblarsen:
> @coliff fixed?

[See above](https://github.com/h5bp/html5-boilerplate/pull/3105#issuecomment-2043580746)

Issue #3109, opened 2024-04-08
Add automated tests on different OS/hardware combinations


roblarsen:
### Prerequisites

- [X] I have [searched](https://github.com/h5bp/html5-boilerplate/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/h5bp/html5-boilerplate/blob/main/.github/CONTRIBUTING.md)

### Proposal

Let's catch things like this broken windows test

### Motivation and context

It's embarrassing!

roblarsen:
closed via #3110

PR #3110, opened 2024-04-09
Windows tests

roblarsen:


roblarsen:
closes #3109 

This is correct behavior. Once one of the two tests in the runner fails, the other one is immediately canceled ("fail fast") 

I still don't know why this test is failing, but we have an automated test for it, so it will never happen again.

coliff:
I'm on Node v18.19.1 on Windows 11 Pro.

`npm run test` is failing:
image

looks like it's related to the `/` path ?

roblarsen:
Failing tests is intentional right now

roblarsen:
> I'm on Node v18.19.1 on Windows 11 Pro.

> looks like it's related to the `/` path ?

Yeah, something changed at some point. It's failing because "\" is not "/" 

v8.0.0 the test worked. I'm going to step through versions of the testing framework to see where it broke.

roblarsen:
And... the tests should be fixed. tl;dr; [node-glob breaking change that I missed being a Codespaces addict](https://github.com/isaacs/node-glob/issues/467)

coliff:
yay! nice one @roblarsen !

image

coliff (approved):

PR #3175, opened 2024-10-19
Added theme color meta tag

Nitesh8957:
### Changes Made
- Added a `<meta name="theme-color" content="#3c790a">` tag to `src/index.html`.

### Reason
This change enhances the user experience on mobile devices by setting a consistent theme color for the browser's address bar.

No related issues were addressed.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,7 @@
   <title></title>
   <link rel="stylesheet" href="css/style.css">
   <meta name="description" content="">
+  <meta name="theme-color" content="#3c790a">
 
   <meta property="og:title" content="">
   <meta property="og:type" content="">


coliff:
Thanks for the PR... If we were to add this we'd need the documentation updating and a build made... (change should be in `dist/` too)

adamaveray:
@coliff The `index.html` file [already has a `theme-color` meta tag](https://github.com/h5bp/html5-boilerplate/blob/e5a7112cff6e894d5e859cf08747e6e2c5e1689e/src/index.html#L22) (added in [this commit from 2018](https://github.com/h5bp/html5-boilerplate/commit/c6b5f62d58d5e206ea77cad115bc755a31b637b2) which coincidentally it looks like you made!), so this PR would add a second entry. I think it can safely be closed.

PR #3281, opened 2025-09-09
dependency updates, rolled up

roblarsen:
Updating multiple dependencies at once, because I've been terrible at dependabot and maintenance in general


coliff:
All good! Build success with no issues.

PR #3408, opened 2026-04-20
fix: all 15 npm dependencies in package in package.json

orbisai0security:
## Summary
Fix high severity security issue in `package.json`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | HIGH |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `package.json:1` |

**Description**: All 15 npm dependencies in package.json use semver range prefixes (^), which permit automatic installation of minor and patch updates. While this is common practice, it means that a compromised upstream package release (e.g., a malicious patch version of archiver or del) would be automatically installed on the next 'npm install' without any explicit developer action. This is a known supply chain attack vector that has been exploited against major organizations.

## Changes
- `package.json`

## Verification
- [x] Build passes
- [x] Scanner re-scan confirms fix
- [x] LLM code review passed

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


coliff:
thanks for the PR but it's best for Dependabot to handle these. Also we need to ensure the lock file is updated too.

PR #3443, opened 2026-06-24
Add proposal guidelines for new projects

roblarsen:
I'm not sure where this should actually live, so this PR might move (or it might not!) but I'm really into this idea (And have been pitched recently) 

Let me know what you think.


roblarsen:
I'm just going to pull this in now. And then we can figure out how to handle it once we reorganize the repo/sites.

PR #3447, opened 2026-06-26
Upgrading eslint.

roblarsen:
I'm not sure why we skipped the eslint upgrade. This appears to work.  Maybe I'm crazy?


coliff:
for good measure - update https://www.npmjs.com/package/eslint-plugin-mocha too?

coliff (approved):

